Issue #5, opened 2014-11-26
Support app arguments


sindresorhus:
Use-case: #4

Something like: `opn('http://todomvc.com', ['chrome', '--incognito']);`

sindresorhus:
fixed in 2.0.0 => https://github.com/sindresorhus/opn/commit/55e6dac9264dc02844b64a870f1d6d40160452a9

Issue #4, opened 2014-11-25
"google-chrome --incognito" won't open, opens normally without flag.


vibegui:
I caught some very strange behaviour and could not think of any possibility on how to solve it.

In [grunt-contrib-connect](https://github.com/gruntjs/grunt-contrib-connect) v0.8, I used "google-chrome --incognito" to develop in incognito mode.

Now that `opn` has been introduced in v0.9, chrome is not launched with this flag.

If I use simply `google-chrome` or `firefox`, it works perfectly.

I logged the `cmd` variable before execFile and it seems OK. 

Any ideas?

PR #6, opened 2014-12-21
completes #5 - support app arguments

tunnckoCore:
Refers to #4 #5 
I think it would work. Im on Linux and work.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -2,29 +2,35 @@
 'use strict';
 var pkg = require('./package.json');
 var opn = require('./');
+var args = process.argv.slice(2);
 
 function help() {
 	console.log([
 		pkg.description,
 		'',
 		'Usage',
-		'  $ opn <file|url> [app]',
+		'  $ opn <file|url> [app] [app arguments]',
 		'',
 		'Example',
 		'  $ opn http://sindresorhus.com',
 		'  $ opn http://sindresorhus.com firefox',
+  		'  $ opn http://sindresorhus.com google-chrome --incognito',
 		'  $ opn unicorn.png'
 	].join('\n'));
 }
 
-if (process.argv.indexOf('--help') !== -1) {
+if (args.indexOf('--help') !== -1) {
 	help();
 	return;
 }
 
-if (process.argv.indexOf('--version') !== -1) {
+if (args.indexOf('--version') !== -1) {
 	console.log(pkg.version);
 	return;
 }
 
-opn(process.argv[2], process.argv[3]);
+opn(args.shift(), [args.shift(), args], function (err) {
+	if (err) {
+		console.error(err);
+	}
+});
--- a/index.js
+++ b/index.js
@@ -12,6 +12,8 @@ module.exports = function (target, app, cb) {
 		app = null;
 	}
 
+	app = app && !Array.isArray(app) ? [app] : app;
+
 	var cmd;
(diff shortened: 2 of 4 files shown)


sindresorhus:
Actually when thinking about it. The Firefox thing shouldn't be a problem. It's up to the user to specify that right. http://superuser.com/questions/413939/how-to-start-a-program-with-command-line-arguments-on-windows-cmd-with-start

So basically: `opn('http://sindresorhus.com', ['firefox', '-option "argument"']);`

Yes, it will not be cross-platform, but there's just no way to achieve that when apps behave differently regarding arguments on different platforms.

I think this is a good middle ground.

Thoughts?

sindresorhus:
@tunnckoCore :arrow_up:

tunnckoCore:
Sorry for the delay.

Yep. But this `['firefox', '-option "argument"']);` is ugly.

As I remember (and see now test cases) it works on linux and os x. I'll check again tests.

And we need some windows guru, hxahxhahxa

sindresorhus:
> is ugly.

I'm surprised you're surprised. It's Windows after all.

tunnckoCore:
ohh yea, right, lol

tunnckoCore:
@sindresorhus 
Hmm.. maybe [cross-spawn](https://github.com/IndigoUnited/node-cross-spawn) should help?

sindresorhus:
How exactly?

arthurvr:
ping @tunnckoCore and @asbjornenge. Anyone of you interested in completing this PR or #8?

tunnckoCore:
@arthurvr how? Windows sucks. We need additions to this PR to handle them correctly, also need setup for windows build testing

asbjornenge:
I could clean up #8 ref. @sindresorhus comments about syntax. But would it then be merged? And if it would work on windows I have no clue. Nor any means of testing.

tunnckoCore:
>  And if it would work on windows I have no clue.

that's the problem, we need testing before merge anything. syntax isnt a problem, it is the easiest thing.

sindresorhus:
Yeah, don't worry about code style. I can just fix that later on. The hold up here is having it work cross-platform and manually testing on all platforms.

tunnckoCore:
> The hold up here is having it work cross-platform and manually testing on all platforms.

mm maybe. and it's not a problem to setup appveyor for testing.

mostly the case is how we should handle these few possible ways of defining arguments in windows and what variant should we limit.

tunnckoCore:
@asbjornenge see https://github.com/sindresorhus/opn/pull/6#issuecomment-67875808

sindresorhus:
> Yep. But this ['firefox', '-option "argument"']); is ugly.

I agree, but we can't really support every app specific way to do args on Windows.

kevva:
@tunnckoCore , if the PR is as complete as it gets as of now, I can test it on Windows.

tunnckoCore:
Mmm nope, imo it is not complete atm, b/c https://github.com/sindresorhus/opn/pull/6#issuecomment-67802183 - just noticed that, they should be changed then we can test.

I will come with some pull-and-commits after dinner.

sindresorhus:
fixed in 2.0.0 => https://github.com/sindresorhus/opn/commit/55e6dac9264dc02844b64a870f1d6d40160452a9

PR #11, opened 2015-06-19
Improve test output

watson:
If these test fails they wouldn't output the errors but just say that they failed. With this commit we use the build in `assert.ifError()` function which shows a better error message if the tests fails

### DIFF
--- a/test.js
+++ b/test.js
@@ -25,7 +25,7 @@ it('should open url in default app', function (cb) {
 	this.timeout(20000);
 
 	opn('http://sindresorhus.com', process.platform === 'darwin' ? cb() : function (err) {
-		assert(!err, err);
+		assert.ifError(err);
 		cb();
 	});
 });
@@ -33,7 +33,7 @@ it('should open url in default app', function (cb) {
 it('should open url in specified app', function (cb) {
 	this.timeout(20000);
 	opn('http://sindresorhus.com', 'firefox', function (err) {
-		assert(!err, err);
+		assert.ifError(err);
 		cb();
 	});
 });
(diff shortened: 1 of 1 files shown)


kevva:
I believe the second argument in `assert()` outputs the error message, doesn't it?

watson:
@kevva Yeah I would have thought so as well, but for some reason I didn't get any output when running the tests even though they failed. Hmm... maybe it's an environment thing. At least when I switched to `ifError()` it worked.

This is my output without using `ifError`:

```
~opn% npm test

> opn@2.0.0 test /Users/watson/code/node_modules/opn
> mocha

  ✓ should open file in default app
  ✓ should open url in default app
  1) should open url in specified app
  2) should open url in specified app with arguments

  2 passing (131ms)
  2 failing
```

This is my output when using `ifError()`:

```
~opn% npm test               

> opn@2.0.0 test /Users/watson/code/node_modules/opn
> mocha

  ✓ should open file in default app
  ✓ should open url in default app
  1) should open url in specified app
  2) should open url in specified app with arguments

  2 passing (89ms)
  2 failing

  1)  should open url in specified app:
     Uncaught Error: Exited with code 1
      at maybeClose (child_process.js:766:16)
      at Socket.<anonymous> (child_process.js:979:11)
      at Pipe.close (net.js:466:12)

  2)  should open url in specified app with arguments:
     Uncaught Error: Exited with code 1
      at maybeClose (child_process.js:766:16)
      at Socket.<anonymous> (child_process.js:979:11)
      at Pipe.close (net.js:466:12)
```

watson:
@kevva can you confirm that you get the error message printet as you expect? I'm running this on node 0.10.38 with mocha 2.2.5.

sindresorhus:
Reopening as the suggested method is more clear.

kevva:
:+1:

Issue #2, opened 2014-06-23
[Question] How to quit opened apps?


sbruchmann:
Hi @sindresorhus,

it seems like there’s currently no way to quit opened apps, since `opn` doesn’t return anything and the `callback` function gets executed once the corresponding app was terminated. Am I missing something?

watson:
I've been toying with the same issue. How about allowing the user to provide a options hash with `{ wait: false }` to override the default behaviour of waiting if `opn` is called with a callback?

In that case it should just call the callback once the child processes returns from opening the URL.

sindresorhus:
@watson :+1: Sounds like a better behaviour than the current way :)

watson:
@sindresorhus Currently the api for the `opn` function is: `opn(target, [app], [callback])`. I think it would be to complicated to both understand and implement if we added an options hash between the `app` and the `callback` argument.

But since the `app` argument is seldom used, couldn't we then just make that part of the options hash? I.e. `opn(target, [options], [callback])`, where options have the following optional properties:

```js
{
  app: "..." // default is platform dependent
  wait: false // default is true
}
```

watson:
That would of cause be a breaking change, which I would rather not introduce... but we could of cause detect if it was given a string and fall back to the old behaviour - or just bump the major version? Let me know what you think

sindresorhus:
`opn(target, [options], [callback])` looks good to me.

I'm fine with breaking backwards compat in a major as I doubt the `app` feature is used much.

watson:
@sindresorhus ^ here you go - let me know what you think :)

PR #13, opened 2015-06-19
Return the child process when calling `opn()`

watson:
This in it self doesn't change any functionality, but will allow people to work directly with the child process how they see fit.

This PR might help with solving #2

### DIFF
--- a/index.js
+++ b/index.js
@@ -89,4 +89,6 @@ module.exports = function (target, app, cb) {
 	} else {
 		cp.unref();
 	}
+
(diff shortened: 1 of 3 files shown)


sindresorhus:
Sure, can you add it to the readme too?

watson:
@sindresorhus Ah yes - good idea. I've updated the PR with the readme changes. Let me know what you think :)

sindresorhus:
Excellent :)

PR #14, opened 2015-06-21
Replace app with options hash to allow disabling waiting for app close

### LINKED ISSUE #2: [Question] How to quit opened apps?

watson:
**WARNING: This is a breaking change!**

Fixes #2

It changes the api of the `opn` function from `opn(target, [app], [callback])` to `opn(target, [options], [callback])`. The use of a custom `app` is still allowed through the new `options.app` property.

It introduces a new `options.wait` boolean (default: `true`) which can be set to `false` if you do not wish the spawned process to keep your program hanging if `opn` is called with a `callback`. The `callback` will in that case be called immediately when the process have finished spawning.

This PR have only implemented `options.wait` for the platforms `darwin` and `win32`. Am I correct in assuming that waiting isn't supported on other platforms?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,35 +1,38 @@
 'use strict';
 var path = require('path');
 var childProcess = require('child_process');
+var objectAssign = require('object-assign');
 
-module.exports = function (target, app, cb) {
+module.exports = function (target, opts, cb) {
 	if (typeof target !== 'string') {
 		throw new Error('Expected a `target`');
 	}
 
-	if (typeof app === 'function') {
-		cb = app;
-		app = null;
+	if (typeof opts === 'function') {
+		cb = opts;
+		opts = null;
 	}
+
+	opts = objectAssign({ wait: true }, opts);
 
 	var cmd;
 	var appArgs;
 	var args = [];
 
-	if (Array.isArray(app)) {
-		appArgs = app.slice(1);
-		app = app[0];
+	if (Array.isArray(opts.app)) {
+		appArgs = opts.app.slice(1);
+		opts.app = opts.app[0];
 	}
 
 	if (process.platform === 'darwin') {
 		cmd = 'open';
 
-		if (cb) {
+		if (cb && opts.wait) {
 			args.push('-W');
 		}
 
-		if (app) {
-			args.push('-a', app);
+		if (opts.app) {
+			args.push('-a', opts.app);
 		}
 
 		if (appArgs) {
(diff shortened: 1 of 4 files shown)


watson:
@sindresorhus how do you normally ensure that the tests run? I guess there is no browser on Circle CI :(

watson:
@sindresorhus let me know if I can do anything to fix the tests, or if this should just be merged as is?

sindresorhus:
Sorry about the delay. Been traveling. I just enabled it on Travis to see if I would be able to set up automated testing, but got busy. Just ignore it. Will review and merge today or tomorrow.

watson:
@sindresorhus no problem - and don't worry... I'm traveling my self, so I know how it is ;)

readme.md
```diff
@@ -56,7 +56,21 @@ The thing you want to open. Can be a URL, file, or executable.
...
+
+An optional options hash.
+
+##### options.wait
+
+Type: `boolean`
+
+Defaults to `true` and means that opn will wait calling the callback (if one is given) until the app have been closed again. If set to `false` the callback will be called immediately after the child process have spawned.
+
+Note that waiting is only possible if the app name is either `darwin` or `win32`.
```
sindresorhus: What does this mean? Did you mean that it will only wait on OS X and Windows?

index.js
```diff
@@ -2,34 +2,38 @@
...
-		cb = app;
-		app = null;
+	if (typeof opts === 'function') {
+		cb = opts;
+		opts = {};
+	} else if (!opts) {
+		opts = {};
 	}
 
+	if (opts.wait !== false) opts.wait = true
```
sindresorhus: Just use https://github.com/sindresorhus/object-assign so not to modify the user supplied options object ;)

readme.md
watson: It's because of this set of if-sentences in the main function:
https://github.com/sindresorhus/opn/blob/a20ea1810c2805b3126134042a8ca9841284aa48/index.js#L24-L65

It seems that the support to actually wait is only available for either `darwin` or `win32` platforms - at least it only seem to be implemented for those two

index.js
watson: Fixed

watson:
@sindresorhus I've updated the PR with your suggestions

readme.md
sindresorhus: I was confused by `app name`. I guess you meant OS name.

But no, it will wait on Linux too, it's just default, so there's no flag.

sindresorhus:
Thanks :)

PR #20, opened 2015-11-04
Bump object-assign version

TrySound:
Just to reduce dep tree.
Also i'd like to suggest you http://greenkeeper.io/ It's pretty cool service for control dependencies.


sindresorhus:
Greenkeeper is super annoying. No way.

PR #23, opened 2016-01-14
promisify

SamVerschueren:
- Replaced `.jshintrc` with XO
- Promisified the API
- AVAfied the tests

Tested all the tests manually and they worked on my Mac. Don't think these changes could break something on Windows or Linux, but off course, you never know for sure. @kevva Are you using Linux? If so, care to give it a go?


SamVerschueren:
Some remarks here. Because it now returns a promise, we can't return the child process unless we would attach it to the promise. Something like

```js
var promise = new Promise(function (resolve, reject) {
    // listen for cp error and close
});

promise.cp = cp;

return promise;
```

Because at the moment, you only receive the child process directly when you are not waiting (wait is `false`) or when the user closes the process when wait is `true`. But when the process is killed, you can't attach listeners to it anymore so there is no point in resolving the promise.

Any feedback is welcome on this one :).

SamVerschueren:
Not sure if `*Required*`should still be added because tje API description describes what parameters are required and which are optional `opn(target, [options])`

watson:
To be honest I'm really sad to see promises taking hold in the Node community :cry:

SamVerschueren:
Not sure why you prefer this:

```js
getData(function(a){  
    getMoreData(a, function(b){
        getMoreData(b, function(c){ 
            getMoreData(c, function(d){ 
                getMoreData(d, function(e){ 
                    ...
                });
            });
        });
    });
});
```
(and now I'm keeping things simple in terms of error handling)

instead of

```js
getData()
    .then(getMoreData)
    .then(getMoreData)
    .then(getMoreData)
    .then(getMoreData)
    .catch(function (err) {
        // handle errors
    });    
```

or in the future with `async`/`await`

```js
try {
    const a = await getData();
    const b = await getMoreData(a);
    const c = await getMoreData(b);
    const d = await getMoreData(c);
} catch (err) {
    // handle errors
}
```

But this indeed is a discussion separate from the PR :).

sindresorhus:
@watson I used to feel exactly like you about promises up until August last year when I took the time to really understand them and use them in lots of modules and projects. Promises simplifies a lot of things when you understand them, but the biggest reason for my choice is eventual async/await syntax, which is so much nicer than the callback mess.

SamVerschueren:
@watson This is an interesting read which I read before using promises. http://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html

PR #29, opened 2016-05-06
put --args [args] after the file argument on OS X

ericmandel:
For the OSX open command, the --args [arg ...] arguments come after the file argument, not before:

````
NAME
     open -- open files and directories

SYNOPSIS
     open [-e] [-t] [-f] [-F] [-W] [-R] [-n] [-g] [-h] [-b bundle_identifier]
          [-a application] file ... [--args arg1 ...]
````

This shows itself as a bug in chrome in the case where chrome is running, but without any active windows.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,11 +31,6 @@ module.exports = function (target, opts) {
 		if (opts.app) {
 			args.push('-a', opts.app);
 		}
-
-		if (appArgs.length > 0) {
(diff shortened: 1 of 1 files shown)


woldie:
I believe this PR will cure a similar failure to launch when app is already running issue I am having with Firefox in a very similar situation on OS X El Capitan

PR #33, opened 2016-08-16
1. Added possibility to open urls on win32 platforms with MS Edge

arghlolrofl:
On Win10 the Edge browser can be launched via:
`(cmd /c) start microsoft-edge:`

Or to launch a url directly:
`(cmd /c) start microsoft-edge:http://www.some.url`

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,6 +66,13 @@ module.exports = function (target, opts) {
 	}
 
 	args.push(target);
+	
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks, but I'm not interested in adding custom workarounds for apps. That's a gigantic slippery slope. If Edge doesn't work with the `start` command you should report it to the Edge developers.

PR #42, opened 2017-03-05
Run process detached from its parent on Linux

GAumala:
Add option 'detached' to allow the opened app to continue running after the
parent process exits.

in Linux and MacOS if this option is true, the child process `stdio` is
set to 'ignore'.

______

The reason for this change is that I use `opn` to open a browser, but I don't want the browser to get closed when the node.js process exits, which is what currently happens, even if I set `wait` to false. The code is a bit messy because this option seems to conflic with `wait`. According to the [docs](https://nodejs.org/api/child_process.html#child_process_options_detached), to spawn a child process so that it can continue after parent exits 2 options are needed:
- set option `detached` true.
- set option `stdio` to 'ignore'.

it might be better add `stdio` option as well instead of just asuming that `detached` true implies the value of `stdio`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,11 +59,9 @@ module.exports = (target, opts) => {
 
 		if (!opts.wait) {
 			// `xdg-open` will block the process unless
-			// stdio is ignored even if it's unref'd
-			cpOpts.stdio = 'ignore';
-			// `xdg-open` will NOT stay running in the background 
-			// after the parent exits unless it is detached from 
-			// the parent AND it ignores the parent's stdio file descriptor 
+			// stdio is ignored and it's detached from the parent
+			// even if it's unref'd
+			cpOpts.stdio = 'ignore'; 
 			cpOpts.detached = true;
 		}
 	}
--- a/index.js
+++ b/index.js
@@ -61,6 +61,10 @@ module.exports = (target, opts) => {
 			// `xdg-open` will block the process unless
 			// stdio is ignored even if it's unref'd
 			cpOpts.stdio = 'ignore';
+			// `xdg-open` will NOT stay running in the background 
+			// after the parent exits unless it is detached from 
+			// the parent AND it ignores the parent's stdio file descriptor 
+			cpOpts.detached = true;
 		}
 	}
 
--- a/readme.md
+++ b/readme.md
@@ -65,7 +65,7 @@ Type: `Object`
 Type: `boolean`<br>
 Default: `true`
 
-Wait for the opened app to exit before fulfilling the promise. If `false` it's fulfilled immediately when opening the app.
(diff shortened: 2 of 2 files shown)


kevva:
What OS are you on?

GAumala:
I'm on Linux. I don't really understand how `wait` should do the same thing. README only says that it controls the moment in which the promise is fulfilled, but it doesn't say anything about what options are set to the spawned process. 

I do recognize that this solution is probably not optimal since increasing the number of options adds up complexity. Do you think that `wait` should also modify `detached` and `stdio` in the spawned process?

kevva:
Well, you're doing [exactly](https://github.com/sindresorhus/opn/blob/master/index.js#L60-L64) the same thing, except that you're waiting for the process to exit before resolving the Promise.

GAumala:
Thanks for the quick reply. I think I finally understood what you meant. I have now changed the code so that when wait is false it sets `stdio` to `'ignore'` and `detached` to `true` on all platforms. No extra options added.  

[The node.js docs](https://nodejs.org/api/child_process.html#child_process_options_detached) show this example for detaching child processes in order to let them run after the parent exits:

``` javascript
const { spawn } = require('child_process');

const subprocess = spawn(process.argv[0], ['child_program.js'], {
  detached: true,
  stdio: 'ignore'
});

subprocess.unref();
```

This is what I want `opn` to do when `wait` is `false`.

I do believe that this issue occurs on all platforms but I'd like to be proven wrong so here's a test plan.

Consider the following script:

``` javascript
const opn = require('opn')

opn('https://fsf.org', { app: 'google-chrome-stable', wait: false}).then(() => {
  console.log('promise fulfilled')
})

while(true);

```

It just opens Google Chrome and with `wait` set to `false` and then gets stuck in an infinite loop. Before you run this make sure that the browser is not already running, we need the script to actually start the browser process not just open a new tab. 

After the page loads, terminate the script with `Ctrl+C`. 

What should happen is that the script exits, but the browser keeps running. 
What actually happens in any recent version of  `opn` is that both the script and the browser exit. 

Why? Because node forwards the interrupt signal to the child process. The only way to avoid it to detach the child process from its parent.

kevva:
I can't reproduce this behaviour in `macOS`, even when using your snippet. When I exit the process the app stays open.

W1lkins:
I can confirm using `macOS Sierra` does not cause this to occur. However, using Arch Linux I can replicate the behaviour described by @GAumala

kevva:
@W1lkins, thanks for testing. @GAumala, could you just move the `detached` option into https://github.com/sindresorhus/opn/blob/master/index.js#L63 and I think this will be good to go.

GAumala:
Yes it seems I've been looking at this at the wrong level. Probably the bug is related to `xdg-open` and it needs to be detached so that the app can continue to run after the node process exits. I have now changed my code to only apply this change on platforms that use `xdg-open`. Thanks for helping guys!

Issue #44, opened 2017-03-12
Incorrectly add "^" before "&" in url query strings


JackCai1206:
I haven't really looked into index.js for this problem, but this doesn't occur on all urls. This only occurs when I have a space in the url I am trying to open. 
To reproduce this issue on windows 10, do `opn('https://www.google.com/#q=12 3&*', {app: 'Chrome'})` and the url in the browser will be `https://www.google.com/#q=12 3^&*`, however, if I do `opn('https://www.google.com/#q=123&*', {app: 'Chrome'})` the url opened in the browser is not modified. 
I noticed there is a regex replacement `target.replace(/&/g, '^&')` in index.js, but im not sure why it doesn't replace all the &s with ^&

PR #47, opened 2017-03-30
Only escape & in Windows when the string has no spaces

arcadiogarcia:
After hitting the bug described in #44  , I played around a little with the spawn method and it seems like Windows' CMD only requires the ^ to escape &s if the string has no spaces, otherwise it will just interpret them as the literal ampersand and the ^ will be added to the string. (I assume that this is because strings with spaces are automatically escaped, to prevent the spaces from causing any trouble).

I added a condition to only escape the & in strings without spaces and it seems to solve those issues.

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,7 +32,10 @@ module.exports = (target, opts) => {
 	} else if (process.platform === 'win32') {
 		cmd = 'cmd';
 		args.push('/c', 'start', '""');
-		target = target.replace(/&/g, '^&');
+		
+		if(target.indexOf(" ") == -1){
+			target = target.replace(/&/g, '^&');
+		}
 
 		if (opts.wait) {
 			args.push('/wait');
--- a/test.js
+++ b/test.js
@@ -40,3 +40,7 @@ test('return the child process when called', async t => {
 	const cp = await m('index.js');
 	t.true('stdout' in cp);
 });
+
+test('open url with spaces and &', async () => {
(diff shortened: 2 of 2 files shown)


sindresorhus:
Thanks for looking into that. Can you find some documentation on it? I'm not comfortable adding something because it "seems" to work. This also needs a test.

arcadiogarcia:
I haven't been able to find any official documentation, but I found this meticulous [article](http://daviddeley.com/autohotkey/parameters/parameters.htm#WINCMDRULES) which explicitly says "Use ^ **outside double quotes** to escape < > & | ^" (bold by me) and "Inside Double Quotes, everything is taken literally ". Given that strings with spaces must be wrapped in double quotes in order to be recognized as a single parameters, that would explain it.

arcadiogarcia:
Just added a test case that failed with the previous code but works now.

sindresorhus:
> Given that strings with spaces must be wrapped in double quotes in order to be recognized as a single parameters, that would explain it.

Not really. Node.js *always* wraps it in double-quotes: https://github.com/nodejs/node/blob/189afc8dc2a4a01e6e97493e167ad6d2db6ff065/lib/child_process.js#L419

---

What we could try is using the `spawn` `{shell: true}` option instead of calling `cmd` directly. It seems it does almost what we do here, just using some more tricks: https://github.com/nodejs/node/blob/189afc8dc2a4a01e6e97493e167ad6d2db6ff065/lib/child_process.js#L417-L420

PR #50, opened 2017-04-14
Add support for Windows Subsystem for Linux

remixz:
With the Windows 10 Creators Update, the Windows Subsystem for Linux now allows for Windows commands to be called from the Linux environment (https://blogs.msdn.microsoft.com/commandline/2017/04/11/windows-10-creators-update-whats-new-in-bashwsl-windows-console/). 

This PR adds support for WSL by using the Windows logic and calling `cmd.exe` while in WSL, instead of `xdg-open`, which doesn't work. Cheers! 👋

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict';
 const path = require('path');
 const childProcess = require('child_process');
+const os = require('os');
 
 module.exports = (target, opts) => {
 	if (typeof target !== 'string') {
@@ -13,6 +14,7 @@ module.exports = (target, opts) => {
 	let appArgs = [];
 	let args = [];
 	const cpOpts = {};
+	const isWSL = (os.type() === 'Linux' && os.release().indexOf('Microsoft') > -1);
 
 	if (Array.isArray(opts.app)) {
 		appArgs = opts.app.slice(1);
@@ -29,8 +31,8 @@ module.exports = (target, opts) => {
 		if (opts.app) {
 			args.push('-a', opts.app);
 		}
-	} else if (process.platform === 'win32') {
-		cmd = 'cmd';
+	} else if (process.platform === 'win32' || isWSL) {
+		cmd = 'cmd' + (isWSL ? '.exe' : '');
 		args.push('/c', 'start', '""');
 		target = target.replace(/&/g, '^&');
 
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 'use strict';
 const path = require('path');
 const childProcess = require('child_process');
-const os = require('os');
+const isWsl = require('is-wsl');
 
 module.exports = (target, opts) => {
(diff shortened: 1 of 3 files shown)


sindresorhus:
I'm willing to add this if you open an issue on https://github.com/Microsoft/BashOnWindows about adding support for `xdg-open`, so it can eventually *just work*.

index.js
```diff
@@ -13,6 +14,7 @@ module.exports = (target, opts) => {
 	let appArgs = [];
 	let args = [];
 	const cpOpts = {};
+	const isWSL = (os.type() === 'Linux' && os.release().indexOf('Microsoft') > -1);
```
sindresorhus: This check is invalid. `os.release()` doesn't return anything with `Microsoft` for me when in WSL. Did you test this?

Instead use https://github.com/sindresorhus/is-wsl

remixz: Yup, I did:



Are you on the latest version of Windows 10 and Ubuntu distro for WSL? It might be possible that the Anniversary update WSL, which uses Ubuntu 14.04, might not include it. Updating to the Creator's Update didn't automatically upgrade the 14.04 to 16.04 either, so that could also be an issue. In any case, I'm fine with using your module if you'd prefer.

remixz:
@sindresorhus I'm pretty sure that it's less an issue of WSL not supporting `xdg-open`, but that `xdg-open` doesn't support WSL. The `xdg-open` script mostly depends on using commands from the various desktop environments, which WSL wouldn't have. The generic opener in `xdg-open` is based on using Linux binary names (such as https://github.com/sindresorhus/opn/blob/master/xdg-open#L825 for browsers), so it wouldn't be able to launch a Windows program.

I think that delegating the launching the `start` command with `cmd.exe` via the interop makes more sense, since then it'll use the default programs that a user has registered in Windows, versus trying to guess the proper executable to use.

index.js
sindresorhus: Yes, I'm on the latest updates. I just did all the updates yesterday. But you're right it didn't install the latest Ubuntu:

```
sindresorhus@SINDRESORHU8D24:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.5 LTS
Release:        14.04
Codename:       trusty
```

I've added a `os.release()` check too for when it's available: https://github.com/sindresorhus/is-wsl/blob/60ea5d57a51ee596cb144ef47187c0476a5a421b/index.js#L10

Can you change to use this module?

sindresorhus:
You're probably right about `xdg-open`. I guess I'm just tired of always having to be forced to add workarounds for Windows.

remixz:
@sindresorhus Sorry for the delay; I've refactored the PR to use the `is-wsl` module now.

Issue #70, opened 2017-09-04
EACCESS on Linux - `xdg-open` does not have execute permission


grinnery:
Installing the package with `npm i opn -S`  (or directly from git repo) does not preserve executable attribute on `xdg-open`, resulting in `EACCESS` when opn-ing URLs.

sindresorhus:
Probably related to https://github.com/npm/npm/issues/18324. Keeping this open until that is resolved.

EvanCarroll:
Why are you even shipping your own version of `xdg-open`. Can't we delegate to `/usr/bin/env` and use the system's version if present?

kevva:
Because not all systems has `xdg-open`. It's probably easier to track down issues by shipping a local version of it.

sindresorhus:
https://github.com/npm/npm/releases/tag/v5.4.1
> This is a very small bug fix release to fix a problem where permissions on installed binaries were being set incorrectly.

PR #71, opened 2017-09-05
Resolves the issues with +x on npm

EvanCarroll:
Works again with linux, resolves the issue with npm disabling -x by sourcing the local `xdg-open` through `/usr/bin/env`


EvanCarroll:
This fixes #70

index.js
```diff
@@ -47,14 +51,15 @@ module.exports = (target, opts) => {
...
+
 		if (opts.app) {
 			cmd = opts.app;
 		} else {
-			cmd = path.join(__dirname, 'xdg-open');
-		}
-
-		if (appArgs.length > 0) {
-			args = args.concat(appArgs);
+			cmd = path.join('/usr', 'bin', 'env');
```
kevva: I'd use something like [which](https://github.com/npm/node-which) here to ensure that we don't use the wrong path.

index.js
```diff
@@ -47,14 +51,15 @@ module.exports = (target, opts) => {
...
+			args.unshift('xdg-open');
```
kevva: Why not `cmd = path.join('/usr', 'bin', 'env', 'xdg-open');`?

index.js
```diff
@@ -47,14 +51,15 @@ module.exports = (target, opts) => {
 			args = args.concat(appArgs);
 		}
 	} else {
+		if (appArgs.length > 0) {
+			args = args.concat(appArgs);
+		}
```
kevva: Why did you move this?

sindresorhus:
Thanks, but I rather just wait on npm to fix this. Looks like a fix there is close: https://github.com/zkat/pacote/pull/117

index.js
EvanCarroll: Because unshifting (pushing to the front) of an array that isn't yet created is a bad idea.

index.js
EvanCarroll: because xdg-open is the argument to `/usr/bin/env` there is no app called `/usr/bin/env/xdg-open`

EvanCarroll:
That's no joy. Shipping out your own copy of `xdg-open` seems like a bad idea to begin with.

PR #72, opened 2017-09-07
Add '/b' flag to hide shell window on Win32

ddantzler:
[Per sindresorhus](https://github.com/sindresorhus/opn/pull/63#issuecomment-327836964), this should be default.

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,7 +32,7 @@ module.exports = (target, opts) => {
 		}
 	} else if (process.platform === 'win32' || isWsl) {
 		cmd = 'cmd' + (isWsl ? '.exe' : '');
-		args.push('/c', 'start', '""');
+		args.push('/c', 'start', '""', '/b');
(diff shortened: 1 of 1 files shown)


kevva:
Could you add a link to the comment just for documentation sake?

index.js
```diff
@@ -32,7 +32,7 @@ module.exports = (target, opts) => {
 		}
 	} else if (process.platform === 'win32' || isWsl) {
 		cmd = 'cmd' + (isWsl ? '.exe' : '');
-		args.push('/c', 'start', '""');
+		args.push('/c', 'start', '""', '/b');
```
kevva: `args.push('/c', '/b', 'start', '""');`

ddantzler: That’s not proper, because /b is a flag for start, not cmd.

kevva: Oh, then ignore my comment :).

kevva:
Thanks 👍

PR #84, opened 2018-02-10
fix: execute xdg-open in directory of opn

nnjpp:
If `/myApp/index.js` imports opn, it should be running `/myApp/node_modules/opn/xdg-open` but is instead looking for `/myApp/xdg-open`

`__dirname` will be the directory of the calling script, not necessarily the directory of opn. This PR looks for xdg-open in `.` instead.

This resolves issues with Linux users getting an ENOENT error as someone mentioned in https://github.com/sindresorhus/opn/issues/56. 

Tested on Ubuntu 17.10 in an Electron + Typescript app.

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,7 +50,7 @@ module.exports = (target, opts) => {
 		if (opts.app) {
 			cmd = opts.app;
 		} else {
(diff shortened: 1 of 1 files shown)


nnjpp:
closing, problem was in my build

PR #88, opened 2018-03-14
Use built-in xdg-open on Android

CrshOverride:
When using `opn` inside of Termux on Android devices (and Chrome OS), the local version of `xdg-open` fails because of architecture issues. In these situations `xdg-open` is available system-wide. This PR updates the `cmd` to use the system-wide `xdg-open` on `android` devices.

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,7 +50,7 @@ module.exports = (target, opts) => {
 		if (opts.app) {
 			cmd = opts.app;
 		} else {
-			cmd = path.join(__dirname, 'xdg-open');
+			cmd = process.platform === 'android' ? 'xdg-open' : path.join(__dirname, 'xdg-open');
 		}
 
 		if (appArgs.length > 0) {


sindresorhus:
> the local version of xdg-open fails because of architecture issues

What architecture issues? How is the system-wide `xdg-open` different from the bundled one?

CrshOverride:
@sindresorhus My apologies. I failed to look at `xdg-open` and assumed the behavior was what I was seeing elsewhere. Architecture has nothing to do with the issue. The file layout when using `termux` on an Android/Chrome OS device is the issue. There is no `/bin/sh` or `/usr/bin/env`. Everything lives in `/data/data/com.termux/files/usr/bin/sh`, etc.

The error I'm seeing is because `/bin/sh` from the shebang can't be found on my device. Using the `xdg-open` from the PATH uses a version with the proper shebang for Android/Chrome OS devices.

PR #74, opened 2017-09-26
Add an option to allow disabling ampersand escaping

MarshallOfSound:
We've found when launching files with a "&" in their filename this escaping breaks the launch.  (Windows throws a could not find file error).  This PR just adds an option to allow disabling the escaping 👍

PR #98, opened 2018-07-06
Fix cannot open url contains `|` on windows

baoti:
- If contains space, **DO NOT** escape;
- otherwise, **DO** escape.
- No matter target is URL, or file path.
- Escape Character: `&` `\` `<` `>` `^` `|`

Detail in [#74 (comment)](https://github.com/sindresorhus/opn/pull/74#issuecomment-403681779).

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,7 +33,7 @@ module.exports = (target, opts) => {
 	} else if (process.platform === 'win32' || isWsl) {
 		cmd = 'cmd' + (isWsl ? '.exe' : '');
 		args.push('/c', 'start', '""', '/b');
-		target = target.replace(/&/g, '^&');
+		target = target.replace(/&/g, '^&').replace(/[|]/g, '^|');
 
 		if (opts.wait) {
 			args.push('/wait');
--- a/index.js
+++ b/index.js
@@ -33,7 +33,9 @@ module.exports = (target, opts) => {
 	} else if (process.platform === 'win32' || isWsl) {
 		cmd = 'cmd' + (isWsl ? '.exe' : '');
 		args.push('/c', 'start', '""', '/b');
-		target = target.replace(/&/g, '^&').replace(/[|]/g, '^|');
+		if (target.indexOf(' ') === -1) {
+			target = target.replace(/([&\\<>^|])/g, '^$1');
+		}
 
 		if (opts.wait) {
 			args.push('/wait');


sindresorhus:
Can you link to some documentation about this?

baoti:
Oh, I found https://ss64.com/nt/syntax-esc.html, look at the **Escape Character** section.

sindresorhus:
What do you think about https://github.com/sindresorhus/opn/pull/74#issuecomment-334714313 ? If it works, it would be better.

baoti:
OK, I‘d like to work on it.

baoti:
@sindresorhus I add many test results in [#74 (comment)](https://github.com/sindresorhus/opn/pull/74#issuecomment-403681779).

sindresorhus:
@baoti Still interested in handling https://github.com/sindresorhus/open/pull/98#issuecomment-402954174?

baoti:
@sindresorhus As [#74 (comment)](https://github.com/sindresorhus/open/pull/74#issuecomment-403681779) said, wrapping it in double quotes doesn't help. What do you want handling?

Cossey:
I think @baoti's [approach mentioned here](https://github.com/sindresorhus/open/pull/98#issue-199644447) would be the best way forward. What an annoying quirk!

sindresorhus:
Alright, let's try out this then:

```
If contains space, DO NOT escape;
otherwise, DO escape.
```

index.js
```diff
@@ -33,7 +33,9 @@ module.exports = (target, opts) => {
 	} else if (process.platform === 'win32' || isWsl) {
 		cmd = 'cmd' + (isWsl ? '.exe' : '');
 		args.push('/c', 'start', '""', '/b');
-		target = target.replace(/&/g, '^&');
+		if (target.indexOf(' ') === -1) {
```
sindresorhus:
```suggestion
		if (!target.includes(' ')) {
```

sindresorhus: Can you also link to https://github.com/sindresorhus/open/pull/74#issuecomment-403681779 in a code comment to clarify what it does and why.

sindresorhus:
@baoti Can you fix the merge conflict?

baoti:
@sindresorhus I got the reason for doing not escape which contains space char. The argument would be wrapped in double quotes. See [source](https://github.com/nodejs/node/blob/728b155870b69a83b806d8bee67aeb0fd0a9b2dd/deps/uv/src/win/process.c#L467) for detail.

We can pass `childProcessOptions.windowsVerbatimArguments = true` to disable it, and always escape target by double quotes.

sindresorhus:
Good find. Let’s do that then.

PR #100, opened 2018-07-30
feat(linux): use system xdg-open when available

### LINKED ISSUE #61: Support pkg
It could be nice if "opn" package support Zeit "pkg" :) .

Currently, building a cross-platform packages with "pkg" don't work well with "opn", it do nothing and browser don't open : 

```
Error: spawn /snapshot/deploy/node_modules/opn/xdg-open ENOENT 
```

I will try some solution and update the issue, but i think "pkg" doesn't work with bash script (xdg-open is not packaged).

axe312ger:
`xdg-open` is already available in many desktop linux distributions. This PR changes the code to detect if it is available on the system and uses that one if detected.

This will fix:
* supplied/bundled xdg-open not compatible with your Linux distribution (Did not encouter this yet but could be a potential issue)
* when bundling your app with tools like https://github.com/zeit/pkg, it is not possible to include the `xdg-open` from this repo in the executable. You need to put the `xdg-open` file in the same directory as your project/app executable, which I'd like to avoid. Should fix #61 for most of the users. (Related: https://github.com/zeit/pkg/issues/213)

I am aware that users on linux without `xdg-open` installed will still have issues, when this lib is bundled via pkg, but I think this risk is acceptable. Especially because when supplying the app via a package manager like apt, a dependency to `xdg-utils` should do the trick.

What do you think @sindresorhus ?

:)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict';
 const path = require('path');
 const childProcess = require('child_process');
+const commandExists = require('command-exists')
 const isWsl = require('is-wsl');
 
 module.exports = (target, opts) => {
@@ -49,8 +50,11 @@ module.exports = (target, opts) => {
 	} else {
 		if (opts.app) {
 			cmd = opts.app;
+		} else if (process.platform === 'android') {
+			cmd = 'xdg-open'
 		} else {
-			cmd = process.platform === 'android' ? 'xdg-open' : path.join(__dirname, 'xdg-open');
+			const xdgOpenInstalled = commandExists.sync('xdg-open')
+			cmd = xdgOpenInstalled ? 'xdg-open' : path.join(__dirname, 'xdg-open')
 		}
 
 		if (appArgs.length > 0) {
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
 		"file"
 	],
 	"dependencies": {
+		"command-exists": "^1.2.7",
 		"is-wsl": "^1.1.0"
 	},
 	"devDependencies": {


sindresorhus:
Many Linux distributions have outdated `xdg-open` versions missing important bug fixes. Bundling `xdg-open` have stability and consistency benefits. I want this package to *just work* out of the box for all Linux users when used with normal Node.js.

> supplied/bundled xdg-open not compatible with your Linux distribution (Did not encouter this yet but could be a potential issue)

I'm not going to change anything based on potential issues.

> when bundling your app with tools like zeit/pkg, it is not possible to include the xdg-open from this repo in the executable. You need to put the xdg-open file in the same directory as your project/app executable, which I'd like to avoid.

I stand by my comment in #61, I should not have to specially work around random tools. This was already marked as fixed in `pkg`, but it seems like the fix was not correct and additional comments on https://github.com/zeit/pkg/issues/207 are ignored. Try opening a new issue on `pkg` about this.

axe312ger:
Alright. Understandable reasons for you package. Thank you for your quick response.

Just want to point out it is not really a bug in pkg, it is more like a inconvenience / UX problem and not that verbose documentation on pkg side.

Will move the detection into our project to use `xdg-open` directly when available.

axe312ger:
We worked around via:

```
  if (['win32', 'darwin'].includes(process.platform)) {
    await opn(URL, {
      wait: false
    })
  } else if (await commandExists('xdg-open')) {
    await execa('xdg-open', [URL])
  } else {
    log(`Unable to open your browser automatically. Please open the following URI in your browser:\n\n${URL}`)
  }
```

zoutepopcorn:
Thanks @axe312ger 

I ve put in on npm https://github.com/zoutepopcorn/opn-pkg

explosivose:
You can also do something like this as a workaround to point to system's xdg-open
```
      let opnOpts: {app?: string | string[]};
      if (['win32', 'darwin'].includes(process.platform)) {
        opnOpts = {};
      } else {
        opnOpts = {app: 'xdg-open'};
      }
      return opn(`http://localhost:${port}/client`, opnOpts);
```

PR #110, opened 2018-11-23
Fix Google Chrome app name

Simplonium:
The current "google chrome" results in an error of windows not being able to find the executable, however "chrome" works. I also tested it with incognito, which also works.


sindresorhus:
https://github.com/sindresorhus/opn/issues/116 would be a good solution for this.

PR #118, opened 2019-03-18
Add support for Windows apps referenced by their WSL paths

### LINKED ISSUE #117: Open using a Windows app referred by its WSL path not supported
When I do work on WSL, oftentimes I need to open files using a Windows binary. One example is opening web URLs, where the `$BROWSER` environment variable is passed in as `app` in the args. This variable is set as a WSL path, e.g. `/mnt/d/Program Files (x86)/Google/Chrome/Application/chrome.exe` instead of `D:\Program Files (x86)\Google\Chrome\Application\chrome.exe`. `cmd.exe` fails to find the application because it does not understand the WSL path.

We can add support to this by detecting if the `app` path is a WSL path and convert it to a Windows path if necessary.

Example use-case:
Currently, the npm package `react-dev-utils` makes a call to `opn`, passing in `$BROWSER`.

stephen-zhao:
Fixes #117

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,12 +4,14 @@ const childProcess = require('child_process');
 const util = require('util');
 const isWsl = require('is-wsl');
 
+const pExecFile = util.promisify(childProcess.execFile);
+
 // Converts a path from WSL format to Windows format
 // e.g. /mnt/c/Program Files/Example/MyApp.exe
 //   => C:\Program Files\Example\MyApp.exe
 const wslToWindowsPath = async path => {
-	const {stdout} = await util.promisify(childProcess.execFile)('wslpath', ['-w', path]);
-	return stdout.toString().trim();
+	const {stdout} = await pExecFile('wslpath', ['-w', path]);
+	return stdout.trim();
 };
 
 module.exports = async (target, options) => {
--- a/index.js
+++ b/index.js
@@ -1,7 +1,16 @@
 'use strict';
 const path = require('path');
 const childProcess = require('child_process');
+const util = require('util');
 const isWsl = require('is-wsl');
+
+// Converts a path from WSL format to Windows format
+// e.g. /mnt/c/Program Files/Example/MyApp.exe
+//   => C:\Program Files\Example\MyApp.exe
+const wslToWindowsPath = async path => {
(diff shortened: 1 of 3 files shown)


stephen-zhao:
Sorry, let me fix these xo errors.

index.js
```diff
@@ -1,8 +1,17 @@
...
 const path = require('path');
 const childProcess = require('child_process');
+const util = require('util');
 const isWsl = require('is-wsl');
 
+// Converts a path from WSL format to Windows format
+// e.g. /mnt/c/Program Files/Example/MyApp.exe
+//   => C:\Program Files\Example\MyApp.exe
+const wslToWindowsPath = async path => {
+	const {stdout} = await util.promisify(childProcess.execFile)('wslpath', ['-w', path]);
```
sindresorhus: Can you use a long flag version of `-w`? For readability.

index.js
```diff
@@ -1,8 +1,17 @@
...
+	return stdout.toString().trim();
```
sindresorhus: `toString()` is moot. It's already a string.

index.js
sindresorhus: Better to promisify at the top-level instead of every time it's called.

```js
const pExecFile = util.promisify(childProcess.execFile);
```

index.js
```diff
@@ -43,6 +52,11 @@ module.exports = async (target, options) => {
 		}
 
 		if (options.app) {
+			if (isWsl && options.app.startsWith('/mnt/')) {
+				const winPath = await wslToWindowsPath(options.app);
+				options.app = winPath;
```
sindresorhus:
```suggestion
				options.app = await wslToWindowsPath(options.app);
```

sindresorhus:
Can you mention in the readme that it now correctly handles WSL mtn paths?

index.js
stephen-zhao: Unfortunately `wslpath` doesn't have any long-form flags.

index.js
stephen-zhao: That fails atomic-assignment check.

stephen-zhao:
@sindresorhus I think it's ready to be merged now?

sindresorhus:
Yup. Looks good :)

PR #131, opened 2019-05-03
fix app option type definition

### LINKED ISSUE #130: Problem with type definition in index.d.ts
In version 6.2.0, there is a type definition on line 33 of index.d.ts which reads:
```
readonly app?: string | readonly string[];
```
(the second `readonly` seems to be the problem)

This causes the following console error:
```
ERROR in /project-path/node_modules/open/index.d.ts
[tsl] ERROR in /project-path/node_modules/open/index.d.ts(33,36)
      TS1005: ';' expected.

ERROR in /project-path/node_modules/open/index.d.ts
[tsl] ERROR in /project-path/node_modules/open/index.d.ts(33,43)
      TS1011: An element access expression should take an argument.

ERROR in /project-path/node_modules/open/index.d.ts
[tsl] ERROR in /project-path/node_modules/open/index.d.ts(35,1)
      TS1128: Declaration or statement expected.
```

If I modify line 33 of index.d.ts from:
```
readonly app?: string | readonly string[];
```
to:
```
readonly app?: string | string[];
```
...then there is no console error.

andyford:
should fix #130 https://github.com/sindresorhus/open/issues/130

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -30,7 +30,7 @@ declare namespace open {
 
 		You may also pass in the app's full path. For example on WSL, this can be `/mnt/c/Program Files (x86)/Google/Chrome/Application/chrome.exe` for the Windows installation of Chrome.
 		*/
-		readonly app?: string | readonly string[];
+		readonly app?: string | string[];
 	}
 }
 


sindresorhus:
https://github.com/sindresorhus/open/issues/130#issuecomment-488989835

PR #134, opened 2019-05-11
Use the system `xdg-open` on Linux when `open` is bundled/packaged

### LINKED ISSUE #133: trying to use /xdg-open when packaged
Been using this in [my vscode extension](https://github.com/marcellourbani/vscode_abap_remote_fs) for a while, and it suddenly broke on linux
Looked into it, issue was adoption of webpack on my side: it does include the extension code but not the packaged xdg-open
Also, __dirname is set to /, as there's no module directory anymore, ending up running /xdg-open which of course doesn't exist.

Easy to work around setting the app option, which I already did, but I think it should default to using the system one in this case. ./xdg-open could make sense, /xdg-open never does

Will submit a PR in a minute

marcellourbani:
Just default to system xdg-open when __dirname = '/' on top of previous cases.

This happens when the code is packaged, as by default xdg-open is not included (and anyway wouldn't be in /)

Fixes #133

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 'use strict';
-const { promisify } = require('util');
+const {promisify} = require('util');
 const path = require('path');
 const childProcess = require('child_process');
 const isWsl = require('is-wsl');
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -73,7 +73,7 @@ module.exports = async (target, options) => {
 		if (options.app) {
 			command = options.app;
 		} else {
-			const useSystemXdgOpen = process.versions.electron || process.platform === 'android';
+			const useSystemXdgOpen = process.versions.electron || process.platform === 'android' || __dirname === '/';
```
sindresorhus: `__dirname === '/'` should be moved to a variable with a readable name.

marcellourbani: Should look better now

PR #138, opened 2019-05-29
Add an option to always use system xdg-open instead of bundled version

deansheather:
Adds a new option `useSystemXdgOpen` which allows for always using the system `xdg-open` command over the included version.

This is neccessary in some environments when the package thinks it can execute the included version when it actually cannot. One example is when the package is included in a binary produced by a node.js binary compiler like [nbin](https://github.com/cdr/nbin).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -31,6 +31,17 @@ declare namespace open {
 		You may also pass in the app's full path. For example on WSL, this can be `/mnt/c/Program Files (x86)/Google/Chrome/Application/chrome.exe` for the Windows installation of Chrome.
 		*/
 		readonly app?: string | readonly string[];
+
+		/**
+		Use system xdg-open instead of the version included.
+
+		When disabled (which is default), the system xdg-open will only be used if running in Electron, if running in Android or if the package is bundled by Webpack.
+
+		This setting is only used when xdg-open will be used.
+
+		@default false
+		*/
+		readonly useSystemXdgOpen?: boolean;
 	}
 }
 
--- a/index.js
+++ b/index.js
@@ -21,6 +21,7 @@ module.exports = async (target, options) => {
 	options = {
 		wait: false,
 		background: false,
+		useSystemXdgOpen: false,
 		...options
 	};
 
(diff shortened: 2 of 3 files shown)


sindresorhus:
> when the package is included in a binary produced by a node.js binary compiler like nbin.

Is there no way of automatically detecting that? I would really prefer to keep the options to a minimum.

deansheather:
There is a way of checking if bundled with `nbin`, by trying to import "nbin" (a package only included in nbin-produced binaries) and seeing if there's an `id` and `mainFile` attribute. I can open a new PR to do that if you want.

It seems that this is a problem that's appeared a few times, however, with the exclusions for Electron and Webpack. I personally think that a configurable option will allow this package to work in many more environments/setups without changes being made to the package.

sindresorhus:
> There is a way of checking if bundled with nbin, by trying to import "nbin" (a package only included in nbin-produced binaries)

Trying to import something that doesn't exist causes problems with Webpack. Any other way? Can you for example check properties on the `module` object?

> I personally think that a configurable option will allow this package to work in 
many more environments/setups without changes being made to the package.

I don't disagree, but the option is useless if you don't directly control `open`, for example, it's used in a dependency of a dependency. That's why automatically handling it would be good in addition to this option.

deansheather:
I don't believe there is any other way of checking if being run inside an nbin binary other than importing `nbin`.

deansheather:
Created a pull request on nbin for `process.versions.nbin` which should allow for checking if open is in an nbin binary: https://github.com/cdr/nbin/pull/9. Once this is merged I will close this PR and make a new one.

notpushkin:
Perhaps a better way would be to check if `path.join(__dirname, 'xdg-open')` actually exists? :) (or just try to run it and fallback to the system one if it fails)

deansheather:
That does seem like a better idea actually. I've created a new PR to check if the bundled `xdg-open` exists and is executable. Closing as the new PR addresses my problem and the owner's concerns with this PR.

PR #140, opened 2019-06-05
Add executable check for local xdg-open

deansheather:
Uses the [executable](https://github.com/kevva/executable) package to ensure that the local/included `xdg-open` is executable. If it's not executable (or there was some other error checking if it was executable, for example `ENOENT`), it will fall back to system `xdg-open` instead.

This allows for use within environments where `open` thought it could run the bundled `xdg-open` but it actually can't, such as in binaries produced by a node.js binary compiler like [nbin](https://github.com/cdr/nbin).

Inspired by https://github.com/sindresorhus/open/pull/138#issuecomment-497332488.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,9 +2,10 @@
 const {promisify} = require('util');
 const path = require('path');
 const childProcess = require('child_process');
-const isExe = require('executable');
+const fs = require('fs');
 const isWsl = require('is-wsl');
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
I would prefer using `fs.access('file', fs.constants.X_OK)`

deansheather:
@sindresorhus Updated the PR, could you review?

PR #146, opened 2019-08-14
Handle URL and paths escaping in Windows

### LINKED ISSUE #17: Strips querystring from URL if it precedes with # in Windows 10 - Chrome
Following code works fine - opens the exact URL in chrome.

``` javascript
var open = require("opn");
open('http://codef0rmer.github.io/eShell/index.html#/1?name=foobar', {app: 'chrome'});
```

However, if Chrome is a default browser and it's not mentioned then both hash and querystrings are stripped.

``` javascript
var open = require("opn");
open('http://codef0rmer.github.io/eShell/index.html#/1?name=foobar');
```

Any idea why `spawn` strips querystring?

### LINKED ISSUE #44: Incorrectly add "^" before "&" in url query strings

### LINKED ISSUE #55: Bug opening file path html file with query param
System: Linux (Ubuntu 14.04)
opn version: 5.1.0
xdg-open version: 1.1.2

File path with query param doesn't work:
```
$ ~/node_modules/opn/xdg-open /some/path/index.html?hi
xdg-open: file '/usr/local/google/home/chenwilliam/lighthouse/plots/charts/index.html?hi' does not exist
```

File path without query param works as expected and opens page in Chrome:
```
$ ~/node_modules/opn/xdg-open /some/path/index.html
Created new window in existing browser session.
```

Regular URL path with query param works:
```
https://www.paulirish.com/?hi
```

Seems related to this issue? https://bugs.freedesktop.org/show_bug.cgi?id=45857

### LINKED ISSUE #77: on windows, can't opn a directory with an ampersand
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/opn/issues/77)

The following line escapes ampersands on windows.
https://github.com/sindresorhus/opn/blob/b9f7e569f5a3d5d520047df3e9f10a42d52fd9d4/index.js#L36

This breaks opening local files and directories where the path contains ampersands.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $30.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($30.00)
### Submitted pull Requests
- [#146 Handle URL and paths escaping in Windows](https://issuehunt.io/r/sindresorhus/opn/pull/146)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/opn/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

### LINKED ISSUE #101: Open local HTML file URL with a query string
When I try and open a local HTML file URL, and I add a query string, then it doesn't work. 

Please make it possible to add query strings when opening local files, like `opn('/file.html?123')`.

### LINKED ISSUE #115: Use the `childProcess.spawn` `shell` option on Windows and fix the quote issue
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/opn/issues/115)

It might simplify some of the ugly command and escaping logic:

https://github.com/sindresorhus/opn/blob/15caf1c276f1251c9cf8c9cebd40c5ec0681a8c9/index.js#L34-L36

And fix the quote issue: https://github.com/sindresorhus/opn/issues/77

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)
### Submitted pull Requests
- [#146 Handle URL and paths escaping in Windows](https://issuehunt.io/r/sindresorhus/opn/pull/146)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/opn/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

herberttn:
This fixes an old problem that has been widely reported but haven't had a simple and concrete solution yet. This is my attempt on making it happen 🙃

**To sum up the problem:** on windows, ampersand (`&`) characters were being escaped with caret (`^`) characters due to windows's cmd interpreting them as a command would break (as in command not found) or would forward a partial argument only (as in URLs with query arguments that contain spaces).

**The solution:** JavaScript's standard `encodeURI` function would solve the problem for URLs, but not for file paths (especially on windows). So we leverage the fact that window's `cmd` interprets a double-quoted argument as a plain text argument just by quoting it (like Node already does). For that to be consistent we need to disable [Node's default quoting and escaping behavior (which is conditional)](https://git.io/fjdem) with its own `windowsVerbatimArguments` flag, and do it ourselves.

Tested in Windows with both Google Chrome and Internet Explorer, using query strings, spaces, pipes, fragments and multiple combinations of all of them.

 - **Fixes #17:** URL fragments
 - **Fixes #44:** URL query strings (ampersand escaping)
 - **Fixes #55:** URL query strings with local HTML files (ampersand escaping)
 - **Fixes #77:** File system path (ampersand escaping)
 - **Fixes #101:** URL query strings with local HTML files (ampersand escaping)
 - **Fixes #115:** Shell option is not needed since windowsVerbatimArguments is now used explicitly
 - **Resolves #74:** File system path (ampersand escaping)
 - **Resolves #98:** URL with pipe characters

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#115: Use the `childProcess.spawn` `shell` option on Windows and fix the quote issue](https://issuehunt.io/repos/18469325/issues/115)
- [#77: on windows, can't opn a directory with an ampersand](https://issuehunt.io/repos/18469325/issues/77)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,7 +55,7 @@ module.exports = async (target, options) => {
 		}
 	} else if (process.platform === 'win32' || isWsl) {
 		command = 'cmd' + (isWsl ? '.exe' : '');
-		cliArguments.push('/c', 'start', '""', '/b');
+		cliArguments.push('/s', '/c', 'start', '""', '/b');
 
 		// Always quoting target allows for URLs/paths to have spaces and unmarked characters, as `cmd.exe` will
 		// interpret them as plain text to be forwarded as one unique argument. Enabling `windowsVerbatimArguments`
--- a/index.js
+++ b/index.js
@@ -26,6 +26,7 @@ module.exports = async (target, options) => {
 	options = {
 		wait: false,
 		background: false,
+		url: false,
 		...options
 	};
 
@@ -37,6 +38,13 @@ module.exports = async (target, options) => {
 	if (Array.isArray(options.app)) {
 		appArguments = options.app.slice(1);
 		options.app = options.app[0];
+	}
+
+	// Encodes the target as if it were an URL. Especially useful to get
+	// double-quotes through the double-quotes on windows caveat, but it
+	// can be used in any platform.
+	if (options.url) {
+		target = encodeURI(target);
 	}
 
 	if (process.platform === 'darwin') {
@@ -61,6 +69,8 @@ module.exports = async (target, options) => {
 		// interpret them as plain text to be forwarded as one unique argument. Enabling `windowsVerbatimArguments`
 		// disables Node.js's default quotes and escapes handling (https://git.io/fjdem).
 		// References: Issues #17, #44, #55, #77, #101 and #115 / Pull requests: #74 and #98
+		//
+		// As a result, all double-quotes are stripped from the `target` and do not get to your desired destination.
 		target = `"${target}"`;
 		childProcessOptions.windowsVerbatimArguments = true;
 
--- a/index.js
+++ b/index.js
@@ -56,7 +56,13 @@ module.exports = async (target, options) => {
 	} else if (process.platform === 'win32' || isWsl) {
 		command = 'cmd' + (isWsl ? '.exe' : '');
 		cliArguments.push('/c', 'start', '""', '/b');
-		target = target.replace(/&/g, '^&');
+
+		// Always quoting target allows for urls/paths to have spaces and unmarked characters, as cmd will
+		// interpret them as plain text to be forwarded as one unique argument. Enabling windowsVerbatimArguments
+		// disables node's default quotes and escapes handling (https://git.io/fjdem).
+		// References: Issues #17, #44, #55, #77, #101 and #115 / Pull requests: #74 and #98
+		target = `"${target}"`;
+		childProcessOptions.windowsVerbatimArguments = true;
 
 		if (options.wait) {
 			cliArguments.push('/wait');
--- a/index.js
+++ b/index.js
@@ -57,9 +57,9 @@ module.exports = async (target, options) => {
 		command = 'cmd' + (isWsl ? '.exe' : '');
 		cliArguments.push('/c', 'start', '""', '/b');
 
-		// Always quoting target allows for urls/paths to have spaces and unmarked characters, as cmd will
-		// interpret them as plain text to be forwarded as one unique argument. Enabling windowsVerbatimArguments
-		// disables node's default quotes and escapes handling (https://git.io/fjdem).
+		// Always quoting target allows for URLs/paths to have spaces and unmarked characters, as `cmd.exe` will
+		// interpret them as plain text to be forwarded as one unique argument. Enabling `windowsVerbatimArguments`
+		// disables Node.js's default quotes and escapes handling (https://git.io/fjdem).
 		// References: Issues #17, #44, #55, #77, #101 and #115 / Pull requests: #74 and #98
 		target = `"${target}"`;
 		childProcessOptions.windowsVerbatimArguments = true;
--- a/index.d.ts
+++ b/index.d.ts
@@ -31,6 +31,17 @@ declare namespace open {
 		You may also pass in the app's full path. For example on WSL, this can be `/mnt/c/Program Files (x86)/Google/Chrome/Application/chrome.exe` for the Windows installation of Chrome.
 		*/
 		readonly app?: string | readonly string[];
+
+		/**
+		Uses `encodeURI` to encode the `target` before executing it.
+
+		The use with targets that are not URLs is not recommended.
+
+		Especially useful when dealing with the [double-quotes on Windows](https://github.com/sindresorhus/open#double-quotes-on-windows) caveat.
+
+		@default false
+		*/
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -56,7 +56,13 @@ module.exports = async (target, options) => {
 	} else if (process.platform === 'win32' || isWsl) {
 		command = 'cmd' + (isWsl ? '.exe' : '');
 		cliArguments.push('/c', 'start', '""', '/b');
-		target = target.replace(/&/g, '^&');
+
+		// Always quoting target allows for urls/paths to have spaces and unmarked characters, as cmd will
+		// interpret them as plain text to be forwarded as one unique argument. Enabling windowsVerbatimArguments
+		// disables node's default quotes and escapes handling (https://git.io/fjdem).
+		// References: Issues #17, #44, #55, #77, #101 and #115 / Pull requests: #74 and #98
+		target = `"${target}"`;
```
sindresorhus: Don't we need to escape any double-quotes in `target`? Can you also add a test for a target containing a double-quote?

sindresorhus:
Sorry for the slow response. I have a lot to catch up on after the summer.

index.js
herberttn: Quick update: I should pick this up again tonight, things have been hectic here too in the past few days.

herberttn: Hey @sindresorhus, I could not make any number of double-quotes go through. Command prompt can only do 1 of 2 things: handle them all **OR** not handle them at all. As per its own documentation:

>*If /C or /K is specified, then the remainder of the command line after the switch is processed as a command line, where the following logic is used to process quote (") characters:*
>
>    1.  *If all of the following conditions are met, then quote characters on the command line are preserved:*
>        - *no /S switch*
>        - *exactly two quote characters*
>        - *no special characters between the two quote characters, where special is one of: &<>()@^|*
>        - *there are one or more whitespace characters between the two quote characters*
>        - *the string between the two quote characters is the name of an executable file.*
>
>    2.  *Otherwise, old behavior is to see if the first character is a quote character and if so, strip the leading character and remove the last quote character on the command line, preserving any text after the last quote character.*

So, since we were already shooting for the second option, I've also added the `/s` switch so we enforce consistency. The final behavior is to handle double-quotes correctly, but that will also mean that all double-quotes in your command/url will magically disappear (in the url cases you could still use `encodeURL` to encode the double-quotes, which will make them get to the destination just fine).

sindresorhus: We can leave it like this for now and if someone complain we can possibly add an option. Can you at least document the behavior so it's clear to people that they need to use `encodeURL` for URLs and that double-quotes be "stripped". It's unlikely that anyone would ever have it in their file paths.

sindresorhus: Actually, how about a `{url: true}` or `{type: 'url'}` option which handles the encoding for the user so they don't have to encode the URL themselves?

herberttn: I've done both :)

Stanzilla:
Does this also fix the problem in https://github.com/sindresorhus/open/pull/98?

herberttn:
> Does this also fix the problem in #98?

Hello @Stanzilla, yes it does!
This one fixes/closes all of the issues/pulls listed in the description!

Since I'm here: Hey @sindresorhus, when you have some spare time, this is ready for another review ;)

sindresorhus:
@herberttn Thanks for your awesome work on this 🙌

sindresorhus:
https://github.com/sindresorhus/open/releases/tag/v7.0.0

herberttn:
> @herberttn Thanks for your awesome work on this 🙌

🎉

baoti:
Maybe need some test for urls contains '%' (e.g. https://cn.bing.com/search?q=%2520 https://cn.bing.com/search?q=%E5%A5%BD). I think `encodeURI` will break url which is already url encoded.

PR #161, opened 2020-01-02
Add option allow positive exit code

### LINKED ISSUE #86: reject on return code > 0
I have a problem with these lines:
```
if (code > 0) {
  reject(new Error('Exited with code ' + code));
  return;
}
```

This assumes that an exit code of > 0 signals an error and other exit codes (0 or negative) signal success.
For one thing: Yes, the exit code is a _signed_ integer and convention says everything other than 0 signals an error.
Second: It's purely a convention that an exit code of 0 signals success, not a standard. Depending on the OS and environment it could just as well signal true, false or "nothing has been processed". An exit code of 5 could signal error or it could signal "I have processed 5 items" or "main() had no return because the programming tutorial said the function signature is 'void main()' and 5 is what happened to be on the stack".

Especially on Windows far too many applications don't follow the 0=success convention and since opn opens a default application we can't know if the application that does get run follows convention, so what is the rejection good for
Essentially as soon as windows is one of your target platforms you have to catch and ignore rejection because there is a good chance it rejects for no reason, but thereby also ignore actual errors reported by the error callback.

Hongarc:
Fixed #86 

I don't know to test with simple exit code so I don't modify `test.js`

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -102,6 +102,13 @@ We do not recommend using it on targets that are not URLs.
 
 Especially useful when dealing with the [double-quotes on Windows](#double-quotes-on-windows) caveat.
 
+##### rejectWithPositive
+
+Type: `boolean`\
+Default: `true`
+
+Reject for positive exit code
+
 ## Caveats
 
 ### Double-quotes on Windows
--- a/index.js
+++ b/index.js
@@ -27,6 +27,7 @@ module.exports = async (target, options) => {
 		wait: false,
 		background: false,
 		url: false,
+		rejectWithPositive: true,
 		...options
 	};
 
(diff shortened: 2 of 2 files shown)


readme.md
```diff
@@ -102,6 +102,13 @@ We do not recommend using it on targets that are not URLs.
 
 Especially useful when dealing with the [double-quotes on Windows](#double-quotes-on-windows) caveat.
 
+##### rejectWithPositive
+
+Type: `boolean`\
+Default: `true`
+
+Reject for positive exit code
```
sindresorhus: The description needs to be a bit clearer about what it does.

sindresorhus:
You need to update index.d.ts too.

readme.md
```diff
@@ -102,6 +102,13 @@ We do not recommend using it on targets that are not URLs.
 
 Especially useful when dealing with the [double-quotes on Windows](#double-quotes-on-windows) caveat.
 
+##### rejectWithPositive
```
sindresorhus: I don't think this option name is clear enough.

Maybe `allowNonzeroExitCode`?

sindresorhus:
// @TanninOne

PR #162, opened 2020-01-02
Fix arg without double-quotes

### LINKED ISSUE #158: can not find program path in windows when specific app option
below code works well:

```javascript
open(resolve(__dirname, "./test.html"), "C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe");
```

But I'm using typescript, as the type definition, I can only use like:

```typescript
open(resolve(__dirname, "./test.html"), {
  app: "C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe"
});
```

I try the version of V6.4.0, that version has not this problem.

package version: V7.0.0
OS version: win10 pro  1909
node version: v12.13.1

Then happen the error like screenshot:
Snipaste_2019-12-12_12-49-50

Hongarc:
Fixes #158

Currently, we have this problem():
image

So I add double-quotes for `options.app`

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,7 +86,7 @@ module.exports = async (target, options) => {
 				options.app = windowsPath;
 			}
 
-			cliArguments.push(options.app);
+			cliArguments.push(`"${options.app}"`);
 		}
 
 		if (appArguments.length > 0) {


sindresorhus:
You need to handle WSL2 too. See: https://github.com/sindresorhus/open/commit/d8a508be1b23c80b53d0089b6f1e487e84f53a19

PR #163, opened 2020-01-09
Added cwd to options

tallev266:


tallev266:
@sindresorhus Hi,
I'd appreciate if you could take a look, I believe this solves the issue when the app needed to run isn't in the global PATH variable.

sindresorhus:
Sorry, but this is not a correct fix.

tallev266:
Thanks for taking a look.

PR #164, opened 2020-01-20
Fix Docker in WSL2

### LINKED ISSUE #157: Incorrect behavior when running in a Docker container in WSL2
Docker containers *do not* have access to the Windows environment, so the `is-wsl` check is not sufficient. The result is that a Docker container on WSL2 fails to run Create React App applications without additional configuration, see logs below.

I was unsure of whether I should make an issue here or in https://github.com/sindresorhus/is-wsl, here's the crux of it:

* Is `is-wsl` supposed to detect if we're running a WSL1/2 kernel? If so, then it is behaving correctly and `open` should likely change.

* Or, is `is-wsl` supposed to detect if we have *access* from a WSL1/WSL2 userland to Windows utilities, such as `cmd.exe`? If so, then `is-wsl` should likely change.

Either way, I have done a bit preliminary research and there are a few reliable ways of detecting if you're running in Docker *and* WSL2, although different container runtimes can do different things:

1. `cat /proc/1/cgroup  | grep docker` exits with status code 0 if running in Docker, because it runs the init process in various cgroups. (We could also check if it contains more than 1 cgroup, or if it contains cgroups that are not the root cgroup, but I'm less sure of the accuracy of those.)

2. The file `/.dockerenv` is injected into Docker containers on WSL2 and absent otherwise, *but* a root user or a distro could theoretically create this file, so it's not reliable.

3. WSL2 distributions do not set the `HOSTNAME` env var, but Docker on WSL2 does.

4. Lastly, I think this may be the best way to detect if we can access Windows: `cat /proc/mounts | awk '{print $3}' | grep drvfs` detects whether or not we have a `drvfs` mount available to us. 🔔 🔔 🔔 this might be the winner.

In which case, I would be happy to make a PR with code like this:

```
const fs = require('fs');

const parseMountString = (mountString) => {
  const [device, mountPoint, fsType, flags, extra] = mountString.split(' ');

  return { device, mountPoint, fsType, flags, extra };
};

const canAccessWindows = fs
  .readFileSync('/proc/mounts')
  .toString()
  .split('\n')
  .map(parseMountString)
  .some(({ fsType }) => fsType === 'drvfs');

console.log(canAccessWindows);
```

See result of running this in a Docker container on WSL2 (left) and in a regular WSL2 distribution (right):

image

Logs showing the error on starting `create-react-app` application, and a [potentially related issue](https://github.com/facebook/create-react-app/issues/7251#issuecomment-564741906):

```logs
> react-app-rewired start

ℹ ｢wds｣: Project is running at http://172.19.0.5/
ℹ ｢wds｣: webpack output is served from /
ℹ ｢wds｣: Content not from webpack is served from /app/public
ℹ ｢wds｣: 404s will fallback to /index.html
Starting the development server...

events.js:174
      throw er; // Unhandled 'error' event
      ^

Error: spawn cmd.exe ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:240:19)
    at onErrorNT (internal/child_process.js:415:16)
    at process._tickCallback (internal/process/next_tick.js:63:19)
Emitted 'error' event at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:246:12)
    at onErrorNT (internal/child_process.js:415:16)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```

mrbrownt:
When running the new experimental docker for desktop (Windows) in WSL2 open assumes that is has access to windows .exe files, which does not work in the context of docker. This uses the is-docker lib to revert to linux behavior in this case but still run things correctly if you're using WSL 1 or 2 outside of docker.

Fixes #157

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ const path = require('path');
 const childProcess = require('child_process');
 const fs = require('fs');
 const isWsl = require('is-wsl');
+const isDocker = require('is-docker');
 
 const pAccess = promisify(fs.access);
 const pExecFile = promisify(childProcess.execFile);
@@ -61,7 +62,7 @@ module.exports = async (target, options) => {
 		if (options.app) {
 			cliArguments.push('-a', options.app);
 		}
-	} else if (process.platform === 'win32' || isWsl) {
+	} else if (process.platform === 'win32' || (isWsl && !isDocker)) {
 		command = 'cmd' + (isWsl ? '.exe' : '');
 		cliArguments.push('/s', '/c', 'start', '""', '/b');
 
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
 		"file"
 	],
 	"dependencies": {
+		"is-docker": "^2.0.0",
 		"is-wsl": "^2.1.0"
 	},
 	"devDependencies": {


pluma:
There's a bug in this PR: `isDocker` is a function, so the check will always return truthy. This completely breaks `open` on WSL as it will fall back to `xdg-open`.

I'll create a PR.

PR #166, opened 2020-01-28
Fix URLs in WSL

### LINKED ISSUE #165: WSL: `open` fails due to quoting issue
## Details

* WSL Version: Ubuntu, Version 2
* Open version: v7.0.0
* Browser version: Firefox 73.0b9

## Steps to reproduce

Open `node`, then:

```
const open = require("open")
open("http://localhost:3000", { wait: false })
```

## What I expect to happen

The browser starts up, and opens localhost:3000

## What actually happens:

image

## Other information

This works with Open v6.3.0.

I suspect commit 7ef15d29a8776737d2d097f02d117c2ec0439d58 is to blame for this issue.

pluma:
Fixes #165.

Verbatim arguments only work when Node is running on Windows, not WSL. Without this fix, start will attempt to open the quoted URL, which will fail.

### DIFF
--- a/index.js
+++ b/index.js
@@ -65,16 +65,18 @@ module.exports = async (target, options) => {
 		command = 'cmd' + (isWsl ? '.exe' : '');
 		cliArguments.push('/s', '/c', 'start', '""', '/b');
 
-		// Always quoting target allows for URLs/paths to have spaces and unmarked characters, as `cmd.exe` will
-		// interpret them as plain text to be forwarded as one unique argument. Enabling `windowsVerbatimArguments`
-		// disables Node.js's default quotes and escapes handling (https://git.io/fjdem).
-		// References:
-		// - Issues #17, #44, #55, #77, #101, #115
-		// - Pull requests: #74, #98
-		//
-		// As a result, all double-quotes are stripped from the `target` and do not get to your desired destination.
-		target = `"${target}"`;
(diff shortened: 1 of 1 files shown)


radar:
@apochilles and I tested this patch out this morning. It works! 

However, we had to pin `isWsl` to 2.1.1 in `package.json`:

```
"is-wsl": "2.1.1"
```

This is because the older version `is-wsl` was not correctly detecting WSL v2 (understandably).

So with that correct version of `is-wsl` and this patch, the browser window does indeed open correctly.

Thanks for the quick patch @pluma!

pluma:
Looks like open already depends on is-wsl `^2.1.0` and 2.1.1 just fixed a bug where `undefined` was returned instead of `false` on Linux (i.e. it shouldn't affect WSL). Any idea why you had to pin it?

radar:
Well, I think another dependency of ours had made `is-wsl` be that version. Putting that line in our `package.json` makes it use the right version. I don't know much about JS package dependencies and how they're resolved, but I'm happy with this solution.

pluma:
Interesting. Normally `^2.1.0` should force the version to match "2.1.0 and all minor and bugfix releases above", so you should only end up with 2.1.0 or 2.1.1 (latest). I'd love to look into what's going on with your package resolution, but this seems liked it'd go too far off-topic for this PR.

pluma:
🎉

PR #168, opened 2020-01-30
Fix Docker in WSL2 for real

pluma:
The module exports a function, not a boolean, so the check for its value would always pass. This breaks WSL (again). We don't need to call the function more than once to find out if we're in docker, so this just calls it right on require.

Fixes the bug introduced in #164.

`<snark>`This wouldn't have happened with TypeScript. ;)`</snark>`

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,7 @@ const path = require('path');
 const childProcess = require('child_process');
 const fs = require('fs');
 const isWsl = require('is-wsl');
-const isDocker = require('is-docker')();
+const isDocker = require('is-docker');
 
 const pAccess = promisify(fs.access);
 const pExecFile = promisify(childProcess.execFile);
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -4,7 +4,7 @@ const path = require('path');
 const childProcess = require('child_process');
 const fs = require('fs');
 const isWsl = require('is-wsl');
-const isDocker = require('is-docker');
+const isDocker = require('is-docker')();
```
sindresorhus: The function should be called when used, not at package load time.

pluma: I'll make amends. My reasoning was that since we're checking the environment it's literally running in, this check should remain stable throughout the module's lifetime. I can see how this might cause problems for testing and such though (although the same would be true for `isWsl`).

pluma:
I've changed the PR so it's invoking on use rather than on import. I've tested the changes locally and WSL now works again.

index.js
sindresorhus: Not just that, but it's an expensive call and you don't want to hurt startup performance.

PR #176, opened 2020-04-27
Add `allowNonzeroExitCode` option

### LINKED ISSUE #86: reject on return code > 0

peterchu999:
fixing issue using #161 as reference 

Fixes #86

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,5 @@
 /// <reference types="node"/>
-import { ChildProcess } from "child_process";
+import {ChildProcess} from 'child_process';
 
 declare namespace open {
 	interface Options {
@@ -44,9 +44,9 @@ declare namespace open {
 		readonly url?: boolean;
 
 		/**
-		Option to allow the opened app resolve the promise with nonzero exit code when `wait` option value `true`.
+		Allow the opened app resolve the promise with nonzero exit code when `wait` option value `true`.
 		
-		We do not recommend using it because by default the app convention for success exit code was 0.
+		We do not recommend using it because by default the app convention for success exit code is 0.
 
 		@default false
 		*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,8 +44,9 @@ declare namespace open {
 		readonly url?: boolean;
 
 		/**
-		Use to allow nonzero exit code. 
-		the convention for success exit code was 0, But in some cases there is an app that doesn't return 0 for the success exit code.
+		Option to allow the opened app resolve the promise with nonzero exit code when `wait` option value `true`.
+		
+		We do not recommend using it because by default the app convention for success exit code was 0.
 
 		@default false
 		*/
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,9 +44,9 @@ declare namespace open {
 		readonly url?: boolean;
 
 		/**
-		Allow the opened app resolve the promise with nonzero exit code when `wait` option value `true`.
+		Allow the opened app to exit with nonzero exit code when the `wait` option is `true`.
 		
-		We do not recommend using it because by default the app convention for success exit code is 0.
+		We do not recommend setting this option. The convention for success is exit code zero.
 
 		@default false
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -42,6 +42,14 @@ declare namespace open {
 		@default false
 		*/
 		readonly url?: boolean;
+
+		/**
+		Use to allow non zero exit code. 
+		the convention for success exit code was 0, But in some cases there is an app that doesn't return 0 for the success exit code.
```
sindresorhus, line 48: Put some more effort into this text.

index.d.ts
```diff
@@ -42,6 +42,14 @@ declare namespace open {
...
+
+		@default true
```
sindresorhus, line 50: It should not be `true` by default.

index.d.ts
```diff
@@ -42,6 +42,14 @@ declare namespace open {
...
+		 */
```
sindresorhus, line 51:
```suggestion
		*/
```

index.d.ts
```diff
@@ -42,6 +42,14 @@ declare namespace open {
 		@default false
 		*/
 		readonly url?: boolean;
+
+		/**
+		Use to allow non zero exit code.
```
sindresorhus, line 47:
```suggestion
		Use to allow nonzero exit code. 
```

peterchu999:
okay i'll try to make it better

index.d.ts
```diff
@@ -42,6 +42,15 @@ declare namespace open {
 		@default false
 		*/
 		readonly url?: boolean;
+
+		/**
+		Option to allow the opened app resolve the promise with nonzero exit code when `wait` option value `true`.
```
sindresorhus, line 47: It's already clear from the context that this is an option.

index.d.ts
```diff
@@ -42,6 +42,15 @@ declare namespace open {
...
+		
+		We do not recommend using it because by default the app convention for success exit code was 0.
```
sindresorhus, line 49: was => is

index.d.ts
```diff
@@ -1,5 +1,5 @@
 /// <reference types="node"/>
-import {ChildProcess} from 'child_process';
+import { ChildProcess } from "child_process";
```
sindresorhus, line 2: Don't do unrelated changes.

peterchu999: sorry this was due to my formatter

readme.md
```diff
@@ -102,6 +102,14 @@ We do not recommend using it on targets that are not URLs.
 
 Especially useful when dealing with the [double-quotes on Windows](#double-quotes-on-windows) caveat.
 
+##### allowNonzeroExitCode
+
+Type: `boolean`\
+Default: `false`
+
+Option to allow the opened app resolve the promise with nonzero exit code when `wait` option value `true`.
```
sindresorhus, line 110: The readme docs should be in sync with the index.d.ts. The text here is different.

peterchu999: so sorry i miss the README

PR #182, opened 2020-05-14
Fix double-encoding of URLs

### LINKED ISSUE #181: url option breaks encoded URLs
The wrong URL will be opened when the `{url: true}` option is set and the URL contains encoded characters. This is particularly problematic when using reserved characters in query parameters (in practice this is common with base64-encoded values). For example:

```javascript
const open = require('open');

const url = new URL('https://httpbin.org/get')
url.searchParams.append('foo', '%');
url.searchParams.append('bar', '&');
url.searchParams.append('baz', '=');

console.log(url.href);
open(url.href, {url: false});
open(url.href, {url: true});
```

The valid string representation of this URL is https://httpbin.org/get?foo=%25&bar=%26&baz=%3D and when opened with `{url: false}` the service will echo the correct parameters:
```
"args": {
  "bar": "&", 
  "baz": "=", 
  "foo": "%"
},
```

However, when opened with `{url: true}` the values get double-encoded:
```
"args": {
  "bar": "%26", 
  "baz": "%3D", 
  "foo": "%25"
},
```

Since query string only gets escaped correctly on WSL when `{url: true}` is set (cf. #171), there's currently no way to open a URL containing reserved characters in the query string on WSL.

I'm thinking that a solution would be to only encode double-quotes, i.e.:

```javascript
if (options.url) {
	target = target.replace(/"/g, '%22');

	if (isWsl) {
		target = target.replace(/&/g, '^&');
	}
}
```

Alternatively:

```javascript
const url = require('url');

// ...

if (options.url) {
	target = url.parse(target).href;

	if (isWsl) {
		target = target.replace(/&/g, '^&');
	}
}
```

ltm:
Fixes #181

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ const {promisify} = require('util');
 const path = require('path');
 const childProcess = require('child_process');
 const fs = require('fs');
+const url = require('url');
 const isWsl = require('is-wsl');
 const isDocker = require('is-docker');
 
@@ -45,7 +46,7 @@ module.exports = async (target, options) => {
 	// double-quotes through the “double-quotes on Windows caveat”, but it
 	// can be used on any platform.
 	if (options.url) {
-		target = encodeURI(target);
+		target = new url.URL(target).href;
 
 		if (isWsl) {
 			target = target.replace(/&/g, '^&');
--- a/index.d.ts
+++ b/index.d.ts
@@ -33,7 +33,7 @@ declare namespace open {
 		readonly app?: string | readonly string[];
 
 		/**
-		Uses `encodeURI` to encode the `target` before executing it.
+		Uses `url.URL` to encode the `target` before executing it.
 
 		The use with targets that are not URLs is not recommended.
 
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -45,7 +46,7 @@ module.exports = async (target, options) => {
 	// double-quotes through the “double-quotes on Windows caveat”, but it
 	// can be used on any platform.
 	if (options.url) {
-		target = encodeURI(target);
+		target = new url.URL(target).href;
```
sindresorhus, line 49: You can just use `URL` as a global.

ltm: The global version of `URL` has been available since Node [v10.0.0](https://nodejs.org/api/url.html#url_class_url). I wasn’t quite sure sure which versions of Node `open` supports, so I went with `url.URL` since it’s been available since Node v7.0.0, v6.13.0.

sindresorhus: Oh my fault. I forgot `open` still targets Node.js 8.

Issue #169, opened 2020-01-31
Error ENOTDIR when using yarn 2's pnp


ridho9:
I was debugging ENOTDIR error when this package is used to open browser by package `webpack-plugin-serve`. After a long debugging I tried to unplug this package (which effectively makes the package has a real node_modules folder) and it works. 

So I guess that the `command` variable is a bit confused when in a pnp environment and cant point to the xdg-open script path.

sindresorhus:
Sorry, I have no intention of spending time on Yarn PnP problems. I would recommend opening an issue on https://github.com/yarnpkg/berry about this incompatibility. They are the ones breaking the whole ecosystem. They should also be the one to fix problems.

ridho9:
Thank you for your response, I guess currently the only solution is just to unplug the package

rbong:
Related: https://github.com/yarnpkg/berry/issues/856

PR #185, opened 2020-07-11
fix: pnp compatibility

merceyz:
**What's the problem this PR addresses?**

`open` currently doesn't work on Linux under Yarn PnP without users manually telling yarn to unplug the package.

Fixes https://github.com/sindresorhus/open/issues/169

**How did you fix it?**

Set `preferUnplugged: true` in the manifest to tell yarn to always unplug(unzip) the package

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
 	"engines": {
 		"node": ">=8"
 	},
+	"preferUnplugged": true,
 	"scripts": {
 		"test": "xo && tsd"
 	},


sindresorhus:
Sorry, I'm not interested in adding Yarn 2 specific properties. It's up to Yarn to maintain compatibility with the ecosystem. You don't see the TypeScript project going around forcing everyone to add types. Instead, they maintain DefinitelyTyped. Yarn could do that too. They could have a compatibility list

arcanis:
> They could have a compatibility list

We actually do. I however find fairly unreasonable to suggest adding Sindresorhus-specific properties in a package manager as a rule rather than an exception.

PR #188, opened 2020-08-08
Use powershell as an alternative URL opener on WSL

### LINKED ISSUE #173: Alternative URL opener on WSL - wslview
Regarding this https://github.com/sindresorhus/open/issues/170 issue, I created an issue in WSL (https://github.com/microsoft/WSL/issues/4981) and discovered this https://github.com/wslutilities/wslu, which is preinstalled in new WSL apps and offer the `wslview` utility, to open URLs.

What do you think about using it, @sindresorhus @pluma ?

tim-stasse:
Fixes #173

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,24 +34,19 @@ module.exports = async (target, options) => {
 	};
 
 	let command;
+	let app;
 	let appArguments = [];
 	const cliArguments = [];
 	const childProcessOptions = {};
 
 	if (Array.isArray(options.app)) {
 		appArguments = options.app.slice(1);
-		options.app = options.app[0];
+		app = options.app[0];
 	}
 
-	// Encodes the target as if it were an URL. Especially useful to get
-	// double-quotes through the “double-quotes on Windows caveat”, but it
-	// can be used on any platform.
+	// Encodes the target as a URL.
 	if (options.url) {
 		target = new url.URL(target).href;
-
-		if (isWsl) {
-			target = target.replace(/&/g, '^&');
-		}
 	}
 
 	if (process.platform === 'darwin') {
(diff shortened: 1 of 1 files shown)


sindresorhus:
Why not use `wslview` as proposed in #173?

tim-stasse:
I wanted to, and actually based this change on it (it uses powershell under the hood). However it doesn't support passing additional arguments to the underlying start command in powershell and therefore wouldn't allow implementing the wait/app options.

index.js
```diff
@@ -65,49 +60,53 @@ module.exports = async (target, options) => {
...
 
-		if (options.app) {
-			cliArguments.push('-a', options.app);
+		if (app) {
+			cliArguments.push('-a', app);
 		}
 	} else if (process.platform === 'win32' || (isWsl && !isDocker())) {
-		command = 'cmd' + (isWsl ? '.exe' : '');
-		cliArguments.push('/s', '/c', 'start', '""', '/b');
+		command = 'powershell' + (isWsl ? '.exe' : '');
```
sindresorhus, line 67: Is it safe to use Powershell for the non-WSL case too? Do all Windows installations have it?

tim-stasse: Yeah it's comes with every windows installation, it's the defacto interpreter for the .NET ecosystem, so it's not at all related to wsl and will work for everyone. I also tested it on both wsl and in windows–using powershell 😅

sindresorhus:
With this change, I guess the `url` option is moot? Since it was added because of `&` escaping problems on Windows.

tim-stasse:
It's up to you really, that would be a breaking API change, but it—theoretically—won't actually break any existing usages if it's removed.

Let me know if you'd prefer to remove it as part of this change or not.

sindresorhus:
We can remove it from the JS code, but it has to stay in the index.d.ts file (with a doc comment that it will be removed in the next major release).

PR #195, opened 2020-09-23
Support WSL configuration where Windows paths are not in PATH

### LINKED ISSUE #173: Alternative URL opener on WSL - wslview

### LINKED ISSUE #186: `cmd.exe` is expected to be available on the PATH in WSL
When I try to use `open` from WSL, it fails to open URLs

```shell
$ npx open-cli http://google.com
npx: installed 82 in 9.186s
events.js:291
      throw er; // Unhandled 'error' event
      ^

Error: spawn cmd.exe ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:268:19)
    at onErrorNT (internal/child_process.js:468:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:274:12)
    at onErrorNT (internal/child_process.js:468:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn cmd.exe',
  path: 'cmd.exe',
  spawnargs: [ '/s', '/c', 'start', '""', '/b', 'http://google.com' ]
}
```

I'm assuming this is because there is no browser or it's otherwise WSL related.

Possibly related: #56, #173
Related: https://github.com/open-wc/open-wc/issues/1763

Pytal:
~~Fixes #173 with the caveat of the `wait` option being ignored as mentioned [here](https://github.com/sindresorhus/open/pull/188#issuecomment-671028351).~~

Fixes #186

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,8 +10,6 @@ declare namespace open {
 
 		On Windows, you have to explicitly specify an app for it to be able to wait.
 
-		On WSL if app is set to `wslview` this option will be ignored.
-
 		@default false
 		*/
 		readonly wait?: boolean;
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,6 +10,8 @@ declare namespace open {
 
 		On Windows, you have to explicitly specify an app for it to be able to wait.
 
+		On WSL if app is set to `wslview` this option will be ignored.
+
 		@default false
 		*/
 		readonly wait?: boolean;
--- a/index.js
+++ b/index.js
@@ -57,14 +57,18 @@ module.exports = async (target, options) => {
 			cliArguments.push('-a', app);
 		}
 	} else if (process.platform === 'win32' || (isWsl && !isDocker())) {
-		command = 'powershell' + (isWsl ? '.exe' : '');
-		cliArguments.push(
-			'-NoProfile',
-			'-NonInteractive',
-			'–ExecutionPolicy',
-			'Bypass',
-			'-EncodedCommand'
-		);
+		if (app === 'wslview') {
+			command = 'wslview';
+		} else {
+			command = 'powershell' + (isWsl ? '.exe' : '');
+			cliArguments.push(
+				'-NoProfile',
+				'-NonInteractive',
+				'–ExecutionPolicy',
+				'Bypass',
+				'-EncodedCommand'
+			);
+		}
 
 		if (!isWsl) {
 			childProcessOptions.windowsVerbatimArguments = true;
@@ -95,8 +99,10 @@ module.exports = async (target, options) => {
 			encodedArguments.push(appArguments.join(','));
 		}
 
-		// Using Base64-encoded command, accepted by PowerShell, to allow special characters.
-		target = Buffer.from(encodedArguments.join(' '), 'utf16le').toString('base64');
+		if (app !== 'wslview') {
+			// Using Base64-encoded command, accepted by PowerShell, to allow special characters.
+			target = Buffer.from(encodedArguments.join(' '), 'utf16le').toString('base64');
+		}
 	} else {
 		if (app) {
 			command = app;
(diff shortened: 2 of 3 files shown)


sindresorhus:
Did you see https://github.com/sindresorhus/open/pull/188#issuecomment-671025350 ?

// @tim-stasse

Pytal:
Yes, I referenced it at the top. I noted the caveat in `index.d.ts`.

sindresorhus:
No, I meant that the existing implementation does the same as `wslview` does under the hood without the caveat, so what's the benefit of this PR?

Pytal:
Right sorry for the misunderstanding.

The use case for this is for users on WSL who do not have Windows paths exposed to WSL as seen in [this create-react-app issue](https://github.com/facebook/create-react-app/issues/9644#issuecomment-695335329) and therefore do not necessarily have `powershell` in PATH.

sindresorhus:
> The use case for this is for users on WSL who do not have Windows paths exposed to WSL as seen in this create-react-app issue and therefore do not necessarily have powershell in PATH.

And there's no way to fix this without using `wslview`? I would explore that option first.

Pytal:
I only know of `wslview` as a utility baked into WSL (on a few distros, most notably Ubuntu) for opening URLs and files in Windows. Do you have any suggestions for an alternative solution where neither `wslview` nor `powershell` is used?

Maybe a path to `powershell` could be hardcoded in when it can't be found in PATH, although this feels hacky and is probably unexpected bahviour for the end-user who has opted to not expose Windows paths.

tim-stasse:
@Pytal I believe the actual issue that needs to be resolved here is #186 

As I mentioned in [#188](https://github.com/sindresorhus/open/pull/188#issuecomment-671028351), `wslview` uses `powershell.exe` under the hood. It works regardless of whether `powershell.exe` is on the path—or not—by programmatically constructing the path to the windows installation folder, and then appending a hard coded path to the powershell binary from there—you can see what I mean by taking a look at lines [111](https://github.com/wslutilities/wslu/blob/8075cbdaf11abcd4c28ebc55ba195bec86a55144/src/wslu-header#L111), [125](https://github.com/wslutilities/wslu/blob/8075cbdaf11abcd4c28ebc55ba195bec86a55144/src/wslu-header#L125), and [154](https://github.com/wslutilities/wslu/blob/8075cbdaf11abcd4c28ebc55ba195bec86a55144/src/wslu-header#L154) from `wslview`'s source code.

Pytal:
@tim-stasse Would the solution mentioned [here](https://github.com/microsoft/WSL/issues/5743#issuecomment-679312603) suffice? Or is it preferable to not use another binary?

tim-stasse:
@Pytal `wslvar` comes from the same set of `wslu` binaries as `wslview`, so yes it should work.

Probably just something like the following should work:

```javascript
command = '$(wslvar systemroot)\System32\WindowsPowerShell\v1.0\powershell.exe'
cliArguments.push(
	'-NoProfile',
	'-NonInteractive',
	'–ExecutionPolicy',
	'Bypass',
	'-EncodedCommand'
);

if (isWsl) {
	command = await wslToWindowsPath(command);
} else {
	childProcessOptions.windowsVerbatimArguments = true;
}
```

Pytal:
@tim-stasse I've made the changes using `wslvar` I'd appreciate your feedback

tim-stasse:
@Pytal Looks good to me 👍

Issue #41, opened 2017-02-28
Incognito Not Working


brianantonelli:
Following your example for spawning an incognito session in Chrome doesn't appear to work anymore. 

```
opn('http://sindresorhus.com', {app: ['google chrome', '--incognito']});
```

The above command, from your readme, opens the window in Chrome, but not in incognito mode.

OSX: `10.12.2`
Chrome: `56.0.2924.87`

sindresorhus:
I think it only works if you don't have Chrome already running. There's a `-n` option for `open` to force a window, but it doesn't seem to work with an URL:

```
open -n -a "Google Chrome" https://sindresorhus.com --args --incognito
```

Just opens an empty incognito window.

rowi1de:
`open -n -a "Google Chrome" --args "https://sindresorhus.com" --incognito`
this will open incognito windows with the url

entozoon:
For me, the fix was `chrome` rather than `'google chrome'`.

[PR here](https://github.com/sindresorhus/opn-cli/pull/9)

brianantonelli:
@entozoon that doesnt work for me.. 

`opn('http://sindresorhus.com', {app: ['chrome', '--incognito']});`

results in:
```
    at ChildProcess.<anonymous> (/Users/monkeymojo/Dev/alks-cli/node_modules/opn/index.js:83:13)
    at Object.onceWrapper (events.js:273:13)
    at ChildProcess.emit (events.js:182:13)
    at maybeClose (internal/child_process.js:961:16)
    at Socket.stream.socket.on (internal/child_process.js:380:11)
    at Socket.emit (events.js:182:13)
    at Pipe._handle.close [as _onclose] (net.js:595:12)
(node:49415) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:49415) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

entozoon:
How frustrating, it works for me in the cli version at least.

I suspect the line where it [pushes opts.app to args](https://github.com/sindresorhus/opn/blob/master/index.js#L43) isn't working correctly for an array of items as it suggests.
Does that sound likely @sindresorhus  ?

cekvenich:
Any updates on this?

[...]

PR #206, opened 2020-11-30
Add `newWindow` option

jebrii:
Very light feature add here. Added an option for opening an application in a new window. This is useful for when the application is already open.

For example, if attempting to open an incognito window in Google Chrome (see #41 ), the proper command in a Mac OS terminal would be:

`open -n -a "Google Chrome" "$url" --args --incognito`

This should solve #41 by allowing users to use this tool to execute that command.

--------------------

EDIT: upon further testing, I've found that the url param in the above command (`"$url"`) must be placed last:
`open -n -a "Google Chrome" --args --incognito "$url"`

I'm adding that change to this PR.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,16 +21,16 @@ declare namespace open {
 
 		@default false
 		*/
-    readonly background?: boolean;
-    
-    /**
-    __macOS only__
+		readonly background?: boolean;
 
-    Open the specified url in a new window. This is useful when the app is expected to already be open.
+		/**
+		__macOS only__
 
-    @default false
-    */
-    readonly newWindow?: boolean;
+		Open the specified url in a new window. This is useful when the app is expected to already be open.
+
+		@default false
+		*/
+		readonly newWindow?: boolean;
 
 		/**
 		Specify the app to open the `target` with, or an array with the app and app arguments.
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,7 +21,16 @@ declare namespace open {
 
 		@default false
 		*/
-		readonly background?: boolean;
+    readonly background?: boolean;
+    
+    /**
+    __macOS only__
+
+    Open the specified url in a new window. This is useful when the app is expected to already be open.
+
+    @default false
+    */
+    readonly newWindow?: boolean;
 
 		/**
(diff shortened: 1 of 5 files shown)


MortenGregersen:
It looks like indention is broken on the lines added.

jebrii:
> It looks like indention is broken on the lines added.

Thanks for bringing this to my attention!

index.d.ts
```diff
@@ -23,6 +23,15 @@ declare namespace open {
 		*/
 		readonly background?: boolean;
 
+		/**
+		__macOS only__
+
+		Open the specified url in a new window. This is useful when the app is expected to already be open.
```
sindresorhus, line 29: `open` doesn't only open URLs. Also `url` => `URL`.

index.js
```diff
@@ -65,9 +66,17 @@ module.exports = async (target, options) => {
 			cliArguments.push('--background');
 		}
 
+		if (options.newWindow) {
+			cliArguments.push('-n');
```
sindresorhus, line 70: Use the long-flag.

index.js
```diff
@@ -65,9 +66,17 @@ module.exports = async (target, options) => {
...
+		}
+
 		if (app) {
 			cliArguments.push('-a', app);
 		}
+
+		if (appArguments.length > 0) {
+			cliArguments.push('--args', ...appArguments);
+		}
```
sindresorhus, line 79: Are you sure moving it here is correct? I think the `target` (https://github.com/sindresorhus/open/pull/206/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R155) is order dependent.

package.json
```diff
@@ -48,6 +48,7 @@
 		"file"
 	],
 	"dependencies": {
+		"child_process": "^1.0.2",
```
sindresorhus, line 51: This is incorrect.

index.d.ts
```diff
@@ -23,6 +23,15 @@ declare namespace open {
...
+
```
sindresorhus, line 30: Mention your specific use-case of opening incognito mode.

sindresorhus (changes requested):

PR #210, opened 2020-12-16
Fix support for WSL when Windows isn't mounted in `/mnt/`

jonaskuske:
In WSL, Windows drives aren't always mounted in `/mnt/` – you can specify the mount location via the `root` key in `wsl.conf`. For this reason I replaced the check that verifies the `app` path starts with `/mnt/` with a check that simply looks for `/` at the start to see if it's a Linux path.

I don't know if the current `/mnt/` check was added to handle some specific edge cases, but I think changing it should be ok.

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,7 +92,7 @@ module.exports = async (target, options) => {
 		}
 
 		if (app) {
-			if (isWsl && app.startsWith('/mnt/')) {
+			if (isWsl && app.startsWith('/')) {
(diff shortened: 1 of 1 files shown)


jonaskuske:
> I don't know if the current `/mnt/` check was added to handle some specific edge cases, but I think changing it should be ok.

Just seen that #210 will add a proper detection for `wslAutoMountRoot` which we could use for the check, but I think this can be merged already anyway

Issue #218, opened 2021-02-03
Incorrect path to PowerShell.exe when mounting point for drive C isn't default in WSL 2


kalexey89:
The #217 PR broke ability to start ReactJS app in dev mode for me.  When trying to start it using `yarn start` command I get an error:
```
events.js:292
      throw er; // Unhandled 'error' event
      ^

Error: spawn /mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:269:19)
    at onErrorNT (internal/child_process.js:465:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)

...
```

As we can see from message - `powershell.exe` not found. That happened because I'm didn't use default mounting point for drive C. My mounting point for C is different - `/c` instead of `/mnt/c`.
But the #217 PR change dynamically determination of mounting point to constant ([see 53 line](https://github.com/sindresorhus/open/blob/5ce319c8482a2c6425e0514e0da1630dae85a986/index.js#L53))
Mounting point should be determined dynamically as it be in v7.3.0.

sindresorhus:
// @anaisbetts

anaisbetts:
I'll see what I can do tomorrow

kalexey89:
@anaisbetts The #219 PR will fix this problem. Please review it. Thanks in advance!

kalexey89:
Fixed by #219 PR.

PR #219, opened 2021-02-03
WSL: Get drives mount point from wsl.conf

kalexey89:
The drives mount point for WSL defined as constant at now. See issue #218. 
This PR will fix it by getting drives mount point dynamically from `wsl.conf` file.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,6 @@ const isWsl = require('is-wsl');
 const isDocker = require('is-docker');
 
 const pAccess = promisify(fs.access);
-const pReadFile = promisify(fs.readFile);
 
 // Path to included `xdg-open`.
 const localXdgOpenPath = path.join(__dirname, 'xdg-open');
@@ -15,26 +14,34 @@ const localXdgOpenPath = path.join(__dirname, 'xdg-open');
 /**
  * Returns the mount point for fixed drives in WSL
  *
- * @async
  * @inner
  * @returns {string} Mount point
  */
-const getWslMountPoint = async () => {
-	const confFile = '/etc/wsl.conf';
+const getWslDrivesMountPoint = (() => {
+	let mountPoint = null;
 
-	if (fs.existsSync(confFile)) {
-		const conf = await pReadFile(confFile, {encoding: 'utf-8'});
-		const value = (/root\s*=\s*(.*)/g.exec(conf)[1] || '').trim();
-
-		if (value) {
-			return value.endsWith('/') ? value : value.concat('/');
+	return function () {
+		if (mountPoint) {
+			// Return memoized mount point value
+			return mountPoint;
 		}
-	}
 
-	// Default value for "root" param
-	// according to https://docs.microsoft.com/en-us/windows/wsl/wsl-config
-	return '/mnt/';
-};
+		const confFile = '/etc/wsl.conf';
+
+		if (!fs.existsSync(confFile)) {
+			// Default value for "root" param
+			// according to https://docs.microsoft.com/en-us/windows/wsl/wsl-config
+			return '/mnt/';
+		}
+
+		const confContent = fs.readFileSync(confFile, {encoding: 'utf-8'});
+
+		mountPoint = (/root\s*=\s*(.*)/g.exec(confContent)[1] || '').trim();
+		mountPoint = mountPoint.endsWith('/') ? mountPoint : mountPoint.concat('/');
+
+		return mountPoint;
+	};
+})();
 
 module.exports = async (target, options) => {
 	if (typeof target !== 'string') {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -7,10 +7,35 @@ const isWsl = require('is-wsl');
...
+ *
+ * @async
+ * @inner
+ * @returns {string} Mount point
+ */
+const getWslMountPoint = async () => {
+	const confFile = '/etc/wsl.conf';
+
+	if (fs.existsSync(confFile)) {
+		const conf = await pReadFile(confFile, {encoding: 'utf-8'});
```
anaisbetts, line 26: Just use fs.readFileSync here, it's super small. If you want to you could also memoize this whole function

index.js
```diff
@@ -7,10 +7,35 @@ const isWsl = require('is-wsl');
...
+ * Returns the mount point for fixed drives in WSL
+ *
+ * @async
+ * @inner
+ * @returns {string} Mount point
+ */
+const getWslMountPoint = async () => {
+	const confFile = '/etc/wsl.conf';
+
+	if (fs.existsSync(confFile)) {
```
anaisbetts, line 25: Maybe flip the condition and early-exit so the entire function isn't in an `if` block

anaisbetts:
Seems reasonable to me /cc @sindresorhus

kalexey89:
@anaisbetts I improved source code of function according to your comments. Please check changes.

index.js
```diff
@@ -11,6 +11,38 @@ const pAccess = promisify(fs.access);
 // Path to included `xdg-open`.
 const localXdgOpenPath = path.join(__dirname, 'xdg-open');
 
+/**
+Get the mount point for fixed drives in WSL.
+
+@inner
+@returns {string} The mount point.
+*/
+const getWslDrivesMountPoint = (() => {
```
sindresorhus, line 20: The `open` function is async, so it should not contain any synchronous IO. That would be unexpected for consumers. You can use `util.promisify` on the `fs` methods.

index.js
```diff
@@ -11,6 +11,38 @@ const pAccess = promisify(fs.access);
...
+	let mountPoint = null;
```
sindresorhus, line 21:
```suggestion
	let mountPoint;
```

I prefer to use `undefined`.

index.js
```diff
@@ -11,6 +11,38 @@ const pAccess = promisify(fs.access);
...
+
+		const confFile = '/etc/wsl.conf';
+
+		if (!fs.existsSync(confFile)) {
+			// Default value for "root" param
+			// according to https://docs.microsoft.com/en-us/windows/wsl/wsl-config
+			return '/mnt/';
+		}
+
+		const confContent = fs.readFileSync(confFile, {encoding: 'utf-8'});
```
sindresorhus, line 37:
```suggestion
		const confContent = fs.readFileSync(confFile, {encoding: 'utf8'});
```

index.js
```diff
@@ -11,6 +11,38 @@ const pAccess = promisify(fs.access);
...
+const getWslDrivesMountPoint = (() => {
+	let mountPoint = null;
+
+	return function () {
+		if (mountPoint) {
+			// Return memoized mount point value
+			return mountPoint;
+		}
+
+		const confFile = '/etc/wsl.conf';
```
sindresorhus, line 29:
```suggestion
		const configFilePath = '/etc/wsl.conf';
```

index.js
```diff
@@ -11,6 +11,38 @@ const pAccess = promisify(fs.access);
...
+		if (!fs.existsSync(confFile)) {
+			// Default value for "root" param
+			// according to https://docs.microsoft.com/en-us/windows/wsl/wsl-config
+			return '/mnt/';
+		}
+
+		const confContent = fs.readFileSync(confFile, {encoding: 'utf-8'});
+
+		mountPoint = (/root\s*=\s*(.*)/g.exec(confContent)[1] || '').trim();
+		mountPoint = mountPoint.endsWith('/') ? mountPoint : mountPoint.concat('/');
```
sindresorhus, line 40: Use `+`, not `concat`.

sindresorhus (changes requested):

index.js
kalexey89: I made this function synchronous according to the previous review (see [comment](https://github.com/sindresorhus/open/pull/219#discussion_r570356185)).

index.js
```diff
@@ -7,10 +7,43 @@ const isWsl = require('is-wsl');
...
+
+	return async function () {
+		if (mountPoint) {
+			// Return memoized mount point value
+			return mountPoint;
+		}
+
+		const configFilePath = '/etc/wsl.conf';
+
+		if (!fs.existsSync(configFilePath)) {
```
sindresorhus, line 32: This should also be async. And should use `fs.access`.

PR #220, opened 2021-02-05
Windows: Fix error with forward slashes in path

### LINKED ISSUE #214: Windows paths with forward-slashes broke in v7.2.0
#188 introduced Powershell and a new method for encoding arguments.  By wrapping the target in, "double quote with double quotes to ensure the inner quotes are passed through",  paths with forward slashes on Windows no longer get reinterpreted as they did before, causing `open` to fail.  

I believe this to also be the cause of sindresorhus/open-cli#16 as `open-cli@6.0.1` works on Windows with a target path like `dist\report.html` but not with `dist/report.html`, unlike earlier versions which worked with both.

Currently, `target` has slashes and backticks added to it before encoding:
```
// target = 'dist/report.html'
encodedArguments.push(`"\`"${target}\`""`); // result: "`"dist/report.html`""

```
... which doesn't actually result in the file being opened.  If instead the target is passed as-is:

```
// target = 'dist/report.html'
encodedArguments.push(target); // result: dist/report.html
```

...then `open` works as expected.

Of course, this isn't a proper solution, because I'm sure those quotes and ticks are being inserted for a purpose, but this hopefully demonstrates the issue for someone else with a better understanding of Powershell.

kalexey89:
This PR will fix #214 issue. The error happens if in the `-FilePath` parameter for `Start-Process` cmdlet contains the escaped quotation character. So just need to not add it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,7 +77,7 @@ module.exports = async (target, options) => {
 			encodedArguments.push(`"\`"${app}\`""`, '-ArgumentList');
 			appArguments.unshift(target);
 		} else {
-			encodedArguments.push(`"\`"${target}\`""`);
+			encodedArguments.push(`"${target}"`);
 		}
 
 		if (appArguments.length > 0) {


sindresorhus:
@anaisbetts Would you be able to review?

kalexey89:
@anaisbetts

anaisbetts:
Seems reasonable to me

sindresorhus:
@kalexey89 It was brought to my attention that this makes it not properly escape the input anymore. For example, if the input is `open('https://foo.com/a\" | Start \"calc')`.

Issue #116, opened 2019-03-16
Add app constants for popular apps


sindresorhus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/opn/issues/116)

App binaries have different names on different platforms. This complicates opening a certain app in a cross-platform manner. We could add some constants for popular apps:

```js
open('https://sindresorhus.com', {app: open.apps.chrome});
```

https://github.com/sindresorhus/opn/blob/5c525b5f4f2472e35ace19d6df746bb355b8060a/test.js#L5-L17

Feedback welcome on how it should look like.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)

### Submitted pull Requests
- [#222 Require Node.js 12, add `.apps` and allow multiple apps to be tried](https://issuehunt.io/r/sindresorhus/opn/pull/222)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/opn/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

IssuehuntBot:
@issuehunt has funded $40.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/18469325/issues/116) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/18469325/issues/116).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/users/44827199/repositories) on IssueHunt to raise funds.

stroncium:
I propose to treat `app`/`app[0]` as value/iterable or function returning value/iterable. Iterables could be just arrays, but it sounds like weird limitation.
then for example chrome app would be defined as
```js
() => {
  if(isWsl) return ['Chrome'];
  switch(process.platform){
    case 'darwin': return 'google chrome canary';
    case 'win32': return 'Chrome';
[...]

PR #222, opened 2021-02-13
Require Node.js 10.17, add `.apps` and allow multiple apps to be tried

Richienb:
```js
const open = require('open');

await open('https://google.com', {
	app: {
		name: open.apps.chrome
	}
});
```

Since we're making a breaking change, I went ahead and bumped it up to Node.js 12. However, I'll roll it back if we're not quite ready.

Fixes: #116

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#116: Add app constants for popular apps](https://issuehunt.io/repos/18469325/issues/116)
---
</details>

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -24,11 +24,11 @@ declare namespace open {
 		readonly background?: boolean;
 
 		/**
-	Specify the `name` of the app to open the `target` with and optionally, app `arguments`. `app` can be an array of apps to try to open and `name` can be an array of app names to try. If each app fails, the last error will be thrown.
+		Specify the `name` of the app to open the `target` with and optionally, app `arguments`. `app` can be an array of apps to try to open and `name` can be an array of app names to try. If each app fails, the last error will be thrown.
 
-	The app name is platform dependent. Don't hard code it in reusable modules. For example, Chrome is `google chrome` on macOS, `google-chrome` on Linux and `chrome` on Windows. If possible, use [`open.apps`](#openapps) which auto-detects the correct binary to use.
+		The app name is platform dependent. Don't hard code it in reusable modules. For example, Chrome is `google chrome` on macOS, `google-chrome` on Linux and `chrome` on Windows. If possible, use [`open.apps`](#openapps) which auto-detects the correct binary to use.
 
-	You may also pass in the app's full path. For example on WSL, this can be `/mnt/c/Program Files (x86)/Google/Chrome/Application/chrome.exe` for the Windows installation of Chrome.
+		You may also pass in the app's full path. For example on WSL, this can be `/mnt/c/Program Files (x86)/Google/Chrome/Application/chrome.exe` for the Windows installation of Chrome.
 		*/
 		readonly app?: App | readonly App[];
 
@@ -47,7 +47,7 @@ declare namespace open {
 
 declare const open: {
 	/**
-	An object containing auto-detected binary names for common apps. Useful to work around cross-platform issues.
+	An object containing auto-detected binary names for common apps. Useful to work around cross-platform differences.
 
 	@example
 	```
--- a/index.d.ts
+++ b/index.d.ts
@@ -64,7 +64,7 @@ declare const open: {
 	});
 	```
 	*/
-	apps: Record<AppName, string | string[]>;
+	apps: Record<open.AppName, string | string[]>;
 
 	/**
 	Open stuff like URLs, files, executables. Cross-platform.
--- a/index.d.ts
+++ b/index.d.ts
@@ -42,7 +42,11 @@ declare namespace open {
 		readonly allowNonzeroExitCode?: boolean;
 	}
 
-	type App = {name: string | readonly string[]; arguments?: readonly string[]};
+	type AppName =
+		| 'chrome'
+		| 'firefox';
+
+	type App = {name: AppName | readonly AppName[]; arguments?: readonly string[]};
 }
 
 declare const open: {
--- a/index.d.ts
+++ b/index.d.ts
@@ -30,42 +30,53 @@ declare namespace open {
 
 		You may also pass in the app's full path. For example on WSL, this can be `/mnt/c/Program Files (x86)/Google/Chrome/Application/chrome.exe` for the Windows installation of Chrome.
 		*/
-		readonly app?: string | readonly string[];
-
-		/**
-		__deprecated__
-
-		This option will be removed in the next major release.
-		*/
-		readonly url?: boolean;
+		readonly app?: App | readonly App[];
 
 		/**
 		Allow the opened app to exit with nonzero exit code when the `wait` option is `true`.
-		
+
 		We do not recommend setting this option. The convention for success is exit code zero.
 
 		@default false
 		*/
 		readonly allowNonzeroExitCode?: boolean;
 	}
+
+	type App = {name: string | readonly string[]; arguments?: readonly string[]};
 }
 
-/**
-Open stuff like URLs, files, executables. Cross-platform.
+declare const open: {
+	/**
+	An object containing auto-detected binary names for common apps. Useful to work around cross-platform issues.
 
-Uses the command `open` on OS X, `start` on Windows and `xdg-open` on other platforms.
+	@example
+	```
+	import open, {apps} from 'open';
 
-There is a caveat for [double-quotes on Windows](https://github.com/sindresorhus/open#double-quotes-on-windows) where all double-quotes are stripped from the `target`.
+	await open('https://google.com', {
+		app: {
+			name: apps.chrome
+		}
+	});
+	```
+	*/
+	apps: Record<string, string>;
 
-@param target - The thing you want to open. Can be a URL, file, or executable. Opens in the default app for the file type. For example, URLs open in your default browser.
-@returns The [spawned child process](https://nodejs.org/api/child_process.html#child_process_class_childprocess). You would normally not need to use this for anything, but it can be useful if you'd like to attach custom event listeners or perform other operations directly on the spawned process.
+	/**
+	Open stuff like URLs, files, executables. Cross-platform.
 
-@example
-```
-import open = require('open');
+	Uses the command `open` on OS X, `start` on Windows and `xdg-open` on other platforms.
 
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -31,26 +34,40 @@ const getWslDrivesMountPoint = (() => {
...
-		mountPoint = (/root\s*=\s*(.*)/g.exec(configContent)[1] || '').trim();
+		mountPoint = (/root\s*=\s*(?<mountPoint>.*)/g.exec(configContent)?.groups?.mountPoint || '').trim();
 		mountPoint = mountPoint.endsWith('/') ? mountPoint : mountPoint + '/';
 
 		return mountPoint;
 	};
 })();
 
-module.exports = async (target, options) => {
+const pTryEach = async (array, mapper) => {
```
Richienb, line 56: Separate module?

index.js
```diff
@@ -1,16 +1,19 @@
...
+import path from 'path';
+import childProcess from 'child_process';
+import {promises as fs} from 'fs';
+import {fileURLToPath} from 'url';
+import isWsl from 'is-wsl';
+import isDocker from 'is-docker';
+import defineLazyProperty from 'define-lazy-prop';
+import AggregateError from 'aggregate-error';
+
+// Node.js ESM doesn't expose __dirname (https://stackoverflow.com/a/50052194/8384910)
```
Richienb, line 10: Can't tell whether this will affect

https://github.com/sindresorhus/open/blob/02bff0124bed4a8841ce8ea92bd44ddd34cd9e2a/index.js#L135-L136

package.json
```diff
@@ -48,13 +49,15 @@
 		"file"
 	],
 	"dependencies": {
-		"is-docker": "^2.0.0",
-		"is-wsl": "^2.1.1"
+		"aggregate-error": "^3.1.0",
+		"define-lazy-prop": "^2.0.0",
```
TrySound, line 53: More dependencies 😢

Richienb: To remove `define-lazy-prop` we'll either need to come up with a new way to provide the data or not cache it. Otherwise, we would just be reinventing the wheel.

To remove `aggregate-error` we'll need to throw a different type of error.

sindresorhus: I don't see it as a problem. `aggregate-error` will eventually go away when we can use the native one.

The other one is tiny: https://bundlephobia.com/result?p=define-lazy-prop@2.0.0

index.js
```diff
@@ -35,29 +38,42 @@ const getWslDrivesMountPoint = (() => {
...
+		} catch {}
 
 		if (!isConfigFileExists) {
 			return defaultMountPoint;
 		}
 
-		const configContent = await pReadFile(configFilePath, {encoding: 'utf8'});
-		const configMountPoint = /root\s*=\s*(.*)/g.exec(configContent);
+		const configContent = await fs.readFile(configFilePath, {encoding: 'utf8'});
+		const configMountPoint = /root\s*=\s*(?<mountPoint>.*)/g.exec(configContent)?.groups?.mountPoint?.trim();
```
sindresorhus, line 50: I think optional chaining requires Node.js 14.

index.js
sindresorhus: How about making it a part of AggregateError?

```js
await AggregateError.tryEach(array);
```

index.js
```diff
@@ -35,29 +38,42 @@ const getWslDrivesMountPoint = (() => {
...
+const pTryEach = async (array, mapper) => {
+	const errors = [];
+
+	for await (const item of array) {
+		try {
+			return await mapper(item);
+		} catch (error) {
+			errors.push(error);
+		}
+	}
```
sindresorhus, line 71: I might be missing something, but I don't see the point of using `for await` here.

```suggestion
	for (const item of array) {
		try {
			await mapper(item);
		} catch (error) {
			errors.push(error);
		}
	}
```

sindresorhus:
Not sure we even need `AggregateError` here. We only really care about the last failure.

package.json
```diff
@@ -10,8 +10,9 @@
 		"email": "sindresorhus@gmail.com",
 		"url": "https://sindresorhus.com"
 	},
+	"type": "module",
```
sindresorhus, line 13: I think it's a bit early to make this a `type` module. I prefer to start with less popular modules. Doing it here this early is just going to cause a lot of support requests.

test.js
```diff
@@ -1,105 +1,68 @@
...
 
 // Tests only checks that opening doesn't return an error
 // it has no way make sure that it actually opened anything.
 
 // These have to be manually verified.
 
-test('open file in default app', async () => {
-	await open('index.js');
+test('open file in default app', async t => {
+	await t.notThrowsAsync(() => open('index.js'));
```
sindresorhus, line 10:
```suggestion
	await t.notThrowsAsync(open('index.js'));
```

sindresorhus (changes requested):

Richienb:
> Not sure we even need `AggregateError` here. We only really care about the last failure.

If we only care about the last error then adding the function to `aggregate-error` won't help.

index.js
```diff
@@ -35,29 +38,42 @@ const getWslDrivesMountPoint = (() => {
...
 	};
 })();
 
-module.exports = async (target, options) => {
+const pTryEach = async (array, mapper) => {
+	const errors = [];
+
+	for await (const item of array) {
+		try {
+			return await mapper(item);
```
Richienb, line 65: The `return` should be kept because we are looking for the value of the first call that succeeds.

Richienb:
Ping

index.d.ts
```diff
@@ -24,48 +24,59 @@ declare namespace open {
 		readonly background?: boolean;
 
 		/**
-		Specify the app to open the `target` with, or an array with the app and app arguments.
+	Specify the `name` of the app to open the `target` with and optionally, app `arguments`. `app` can be an array of apps to try to open and `name` can be an array of app names to try. If each app fails, the last error will be thrown.
```
sindresorhus, line 27: Incorrect indentation amount.

index.js
```diff
@@ -193,3 +214,37 @@ module.exports = async (target, options) => {
...
+	['darwin', 'firefox'],
+	['win32', 'C:\\Program Files\\Mozilla Firefox\\firefox.exe'],
+	['linux', 'firefox']
+]), {
+	wsl: '/mnt/c/Program Files/Mozilla Firefox/firefox.exe'
+}));
+
+open.apps = apps;
+
+export default open;
```
sindresorhus, line 250: Still ESM

package.json
```diff
@@ -11,7 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"engines": {
-		"node": ">=8"
+		"node": "^10.17 || >=11.14"
```
sindresorhus, line 14:
```suggestion
		"node": "^10.17"
```

11 is not a supported version anyway.

readme.md
```diff
@@ -84,32 +82,36 @@ Do not bring the app to the foreground.
...
 
-Uses `URL` to encode the target before executing it.<br>
-We do not recommend using it on targets that are not URLs.
+Allow the opened app to exit with nonzero exit code when the `wait` option is `true`.
 
-Especially useful when dealing with the [double-quotes on Windows](#double-quotes-on-windows) caveat.
+We do not recommend setting this option. The convention for success is exit code zero.
 
-##### allowNonzeroExitCode
+### open.apps
```
sindresorhus, line 102: The readme needs to document what constants we expose.

index.d.ts
```diff
@@ -24,48 +24,59 @@ declare namespace open {
...
 	}
+
+	type App = {name: string | readonly string[]; arguments?: readonly string[]};
 }
 
-/**
-Open stuff like URLs, files, executables. Cross-platform.
+declare const open: {
+	/**
+	An object containing auto-detected binary names for common apps. Useful to work around cross-platform issues.
```
sindresorhus, line 50: Not "issues", rather "differences"

readme.md
```diff
@@ -28,20 +28,18 @@ $ npm install open
...
 
-	// Opens the URL in the default browser.
-	await open('https://sindresorhus.com');
+// Opens the URL in the default browser.
+await open('https://sindresorhus.com');
 
-	// Opens the URL in a specified browser.
-	await open('https://sindresorhus.com', {app: 'firefox'});
+// Opens the URL in a specified browser.
+await open('https://sindresorhus.com', {app: 'firefox'});
```
sindresorhus, line 39: The `app` option uses outdated syntax.

sindresorhus (changes requested):

index.d.ts
```diff
@@ -24,65 +24,80 @@ declare namespace open {
...
+	import open = require('open');
+
+	await open('https://google.com', {
+		app: {
+			name: open.apps.chrome
+		}
+	});
+	```
+	*/
+	apps: Record<open.AppName, string | string[]>;
```
sindresorhus, line 67: @Richienb I noticed after merging that this is not correct.

index.js
```diff
@@ -193,3 +214,37 @@ module.exports = async (target, options) => {
...
+}
+
+const apps = {};
+
+defineLazyProperty(apps, 'chrome', () => detectPlatformBinary(new Map([
+	['darwin', 'google chrome canary'],
+	['win32', 'Chrome'],
+	['linux', ['google-chrome', 'google-chrome-stable']]
+]), {
+	wsl: '/mnt/c/Program Files (x86)/Google/Chrome/Application/chrome.exe'
```
sindresorhus, line 237: What if the system is x64?

package.json
```diff
@@ -11,7 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"engines": {
-		"node": ">=8"
+		"node": "^10.17"
```
sindresorhus, line 14: This is also incorrect.

```suggestion
		"node": ">=10.17"
```

index.d.ts
Richienb: @sindresorhus Ok, I'll fix

Richienb: As it turns out, it **is** correct.

Issue #236, opened 2021-03-21
Closes cmd/powershell when opening url


nutgaard:
Lib version: 8.0.3
Nodejs: 15.11.0
Window: Win 10 build 19042.804

Steps to reproduce;
```javascript
const open = require('open');
open('http://github.com');
```
Then: `node index.js` 

Expected behaviour;
1. Opens url in default browser
2. Cmd/powershell process is kept running.

Actual behaviour:
1. Opens url in default browser
2. Kills cmd/powershell process

sindresorhus:
Is this a new issue in 8.0.3? If not, can you test some older versions?

nutgaard:
Just tested it, and was on my way to update the issue. :D 

`open@8.0.2` works as expected.

sindresorhus:
// @kykint

kykint:
According to your code  - you don't use "wait: true" option. Without this option powershell shouldn't keep running.

nutgaard:
I could understand it exiting out of the node-script sure. 
But killing the entire terminal-process seems a bit like over-reaching to be honest, but maybe that has been the expected behaviour all along?
Anyways, I'll test it out with `{wait: true}`.

As a side-note, shouldn't this have been a major-version bump? Pretty drastic change in behaviour for patch-bump.

nutgaard:
Done some more testing outside of my project to easier isolate and test. 
By the looks of it then both of the snipplets below kills the cmd process when using `open@8.0.3`;

```
const open = require('open');
open("http://github.com", { wait: true });
```

```
const open = require('open');

(async () => {
  await open("http://github.com", { wait: true });
})();
```

kykint:
I understood the problem. cmd window is not killed - It's just hidden. Will try to fix it

kykint:
Maybe the best way is just add windows-only option, like "hiddenTerminal". Because there is another situation, when node.js application is started not from cmd. Then powershell creates one more window and it appears on top of all other applications and it is also not good. I can't find better solution yet.

nutgaard:
I don't know which situations you're referring to (electron-app? pkg-bundled app?). 
But unless there is a reliable way of identifying those situations, then that seems like a better solution.

kykint:
I tried to find a way, how to determine environment, but didn't find :( So, I've created a PR #238 with an addition option.

adamtay:
[...]

PR #238, opened 2021-03-21
Fix issue #236. Additional option "hiddenTerminal"

kykint:


sindresorhus:
I don't add options lightly. Please explain why the behavior happens in the first place. If there's a way to solve this without an option, I would prefer that.

qm3ster:
Sure seemed like this 👉 https://github.com/sindresorhus/open/pull/235 👈  option in the call to powershell was added lightly 😉
What happens is that running `powershell.exe –WindowStyle Hidden` from a terminal, where a new Powershell window wouldn't spawn anyway, instead reconfigures the window of the host to making it invisible. The same happens even if nodejs is a middleman.

This is a very confusing behavior, and for a while both the person that asked for my help with this issue and the creator of https://github.com/sindresorhus/open/issues/236 thought the host is being terminated.

sindresorhus:
Alright, PR welcome to revert that commit then.

qm3ster:
But then a PowerShell window will pop up and scare the users when used from a graphical application binary.

met583:
what steps should i take to prevent the terminal from being hidden?  Fix (lookup)?

qm3ster:
@met583 you can use `"open": "8.0.2"` (not `"open": "^8.0.2"` or `"open": "8.0.5"`) until this is merged.

met583:
> @ met583 puoi usare `"open": "8.0.2"`(not `"open": "^8.0.2"`or `"open": "8.0.5"`) fino a quando non viene unito.
> Puoi ottenerlo con `npm i -E open@8.0.2`( `-E`per "esatto")

Thanks, where should I enter that command?  In the cdm "open" is not recognized as an internal or external command.  How do I proceed?

PR #247, opened 2021-05-05
WSL: Ignore commented mount point lines in wsl.conf

### LINKED ISSUE #246: WSL drive C mounting point incorrectly detected from commented lines in wsl.conf
I was having an issue with incorrect detection of the drive C mount point in WSL2 as mentioned in issue #218, despite this having been fixed in PR #219. It was detecting it according to my old configuration even though I'd changed it back to `/mnt/c` by commenting out the lines in wsl.conf. After removing the commented lines it worked as expected.

Looking at [this commit](https://github.com/sindresorhus/open/pull/219/commits/9a89c24761db9a2275c32ac62638c01199509baf) in the PR, the problem seems to be the regex which isn't excluding lines that start with `#`. The problem does not occur if I add a negative lookbehind:

`const configMountPoint = /(?<!#.*)root\s*=\s*(?<mountPoint>.*)/g.exec(configContent);`

haroonrafi:
The mount point is detected by reading wsl.conf (PR #219), but the regex does not ignore commented lines (see #246). This PR adds a negative lookbehind for lines starting with a hash character.

Fixes #246

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,7 +42,7 @@ const getWslDrivesMountPoint = (() => {
 		}
 
 		const configContent = await fs.readFile(configFilePath, {encoding: 'utf8'});
-		const configMountPoint = /root\s*=\s*(?<mountPoint>.*)/g.exec(configContent);
+		const configMountPoint = /(?<!#.*)root\s*=\s*(?<mountPoint>.*)/g.exec(configContent);
 
 		if (!configMountPoint) {
 			return defaultMountPoint;


index.js
```diff
@@ -42,7 +42,7 @@ const getWslDrivesMountPoint = (() => {
 		}
 
 		const configContent = await fs.readFile(configFilePath, {encoding: 'utf8'});
-		const configMountPoint = /root\s*=\s*(?<mountPoint>.*)/g.exec(configContent);
+		const configMountPoint = /(?<!#.*)root\s*=\s*(?<mountPoint>.*)/g.exec(configContent);
```
sindresorhus, line 45: The regex results in an error: https://regex101.com/r/4pJXHS/1

haroonrafi: I'm not sure I see the error, has the regex101 test string changed? I just see `#root` and `root` below it which doesn't match, but that's no different to the original regex because it expects `=` after root. If I change the second line to `root=/` it finds the match group. 

I've tested this change locally with a `#root` comment directly above the actual mount point setting and it worked as expected.

sindresorhus: Ok. It turns out I was looking at it with Safari and Safari doesn't support negative lookbehind.

PR #253, opened 2021-05-21
Add `newInstance` option for macOS

Kudo:
The open command in macOS will try to reuse existing instance by default.
Adding `open -n` will create a new instance no matter if there is existing instance or not.
For Windows and Linux, the behavior are to create new instances anyway.
The PR proposes to add a macOS only `newInstance` option and by default it's false.

Not sure if the change making sense or not. feel free to comment for better solution.

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,6 +77,7 @@ const open = async (target, options) => {
 	options = {
 		wait: false,
 		background: false,
+		newInstance: false,
 		allowNonzeroExitCode: false,
 		...options
 	};
@@ -113,6 +114,10 @@ const open = async (target, options) => {
 
 		if (options.background) {
 			cliArguments.push('--background');
+		}
+
+		if (options.newInstance) {
+			cliArguments.push('-n');
 		}
 
 		if (app) {
--- a/index.js
+++ b/index.js
@@ -117,7 +117,7 @@ const open = async (target, options) => {
 		}
 
 		if (options.newInstance) {
-			cliArguments.push('-n');
+			cliArguments.push('--new');
 		}
 
 		if (app) {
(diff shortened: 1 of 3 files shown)


Kudo:
thanks @sindresorhus for landing edge support. any chance for you to review this?
please feel free to let me know if you have any concern for this.

index.js
```diff
@@ -115,6 +116,10 @@ const open = async (target, options) => {
 			cliArguments.push('--background');
 		}
 
+		if (options.newInstance) {
+			cliArguments.push('-n');
```
sindresorhus, line 120: Use the long-flag.

readme.md
```diff
@@ -80,6 +80,15 @@ Default: `false`
 
 Do not bring the app to the foreground.
 
+##### newInstance <sup>(macOS only)</sup>
+
+Type: `boolean`\
+Default: `false`
+
+Open a new instance of the application(s) even if one is already running.
```
sindresorhus, line 88:
```suggestion
Open a new instance of the app even it's already running.
```

readme.md
```diff
@@ -80,6 +80,15 @@ Default: `false`
...
+
+On other platforms, this is always to open a new instance.
```
sindresorhus, line 90: This needs to be better written.

sindresorhus (changes requested):

readme.md
Kudo: Updated to `For other platforms, this value is always to be true.`
is this better?

Kudo:
updated changes based on your comment, please help to review again.
thank you so much @sindresorhus

Kudo:
@sindresorhus Thank you for the great revise and quick review 👍

PR #258, opened 2021-07-03
docs: add usage of data URLs

### LINKED ISSUE #208: support new format data:text/html
`data:text/html;charset=utf-8,${html}` can be render by chrome.
maybe it can be supported.

eg. 
> "C:\Program Files (x86)\Google\Chrome\Application\chrome.exe" --args --new-window --app="data:text/html,\<h1\>Hello\</h1\>"

g8up:
fixed #208


sindresorhus:
This is not the way I would recommend doing this. It should use the URL argument. If that's not working, it should be fixed.

PR #267, opened 2021-11-05
Add windowsHide option for child_process.spawn

leekafai:
Add windowsHide option for child_process.spawn, aviod scaring people by the pop up PowerShell window
[#238(comment)](https://github.com/sindresorhus/open/pull/238#issuecomment-822023811)


sindresorhus:
There's a lot of unrelated style changes here. There also does not seem to be much demand for this. Very few have requested it. So I'm going to pass on this to keep the complexity down.

Issue #266, opened 2021-10-29
To open url on default browser with incognito mode on cross platforms 


hectorvp:
How do I open default browser  in incognito mode for cross platforms.
If I provide app explicitly, I won't be able to ensure whether that app (browser) is available or not.

```javascript
await open("https://www.google.com", { app: { name: open.apps.chrome , arguments: ["--incognito"] } });
```
In above example, it may be possible that chrome isn't installed and hence not default on linux machine.

Is something like this available anywhere??

```javascript
await open("https://www.google.com", {target: 'browser', 'flags':{incognito: true}});
```

sindresorhus:
I suggest we add two more built-in constants:

- `open.apps.browser`
- `open.apps.browserPrivate` (this one would pass the correct flag)

PR welcome. But it's expected that you know what you're doing, and that you add tests, docs, and types.

PR #281, opened 2022-03-18
Fixed problem with open in macOS Monterey

piotar:
Fixed opening problem with passed arguments in macOS Monterey

### DIFF
--- a/index.js
+++ b/index.js
@@ -194,12 +194,12 @@ const baseOpen = async options => {
 		}
 	}
 
-	if (options.target) {
-		cliArguments.push(options.target);
-	}
-
 	if (platform === 'darwin' && appArguments.length > 0) {
 		cliArguments.push('--args', ...appArguments);
+	}
+
+	if (options.target) {
+		cliArguments.push(options.target);
 	}
 
 	const subprocess = childProcess.spawn(command, cliArguments, childProcessOptions);


sindresorhus:
Can you elaborate?

Because the `open` help on Monterey documents this order:

> Usage: open [-e] [-t] [-f] [-W] [-R] [-n] [-g] [-h] [-s <partial SDK name>][-b <bundle identifier>] [-a <application>] [-u URL] [filenames] [--args arguments]

piotar:
You have absolutely right about syntax command. But when we want pass additional arguments, the browser will open the target in tab, but command open will ignore passed arguments.

Probably, when passing additional arguments, the target should be passed as the argument, not as the `file` argument in the command.

Some cases (before each case, browser should be permanently closed):
```
open -a firefox "http://example.com"
```
This command open 2 instance of firefox
- new instance
- instance with opened target


```
open -a firefox "http://example.com" --args -private-window
```
This command open 2 instance of firefox
- target will be open in browser without private mode
- new instance with private mode

```
open -a firefox --args -private-window "http://example.com"
```
This command will works as expected (one instance with private mode with correct target)

sindresorhus:
The last command works as the URL is the value for the `-private-window` flag: https://wiki.mozilla.org/Firefox/CommandLineOptions#macOS

Closing as your PR is incorrect, but I suggest commenting your findings in https://github.com/sindresorhus/open/issues/266 as we plan to add native support for opening in private mode.

PR #284, opened 2022-05-24
Don't use empty catch blocks please

FilipParyz:
I suppose a change to not use an empty catch block and at least print an error message.

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,7 +35,9 @@ const getWslDrivesMountPoint = (() => {
 		try {
 			await fs.access(configFilePath, fsConstants.F_OK);
 			isConfigFileExists = true;
-		} catch {}
+		} catch (error) {
+			console.error("Access error in npm module 'open'.");
+		}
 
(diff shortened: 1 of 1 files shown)


FilipParyz:
obraz
While using [PKG](https://www.npmjs.com/package/pkg) to package my node.js app using 'open' module, I'm running into a problem as shown on the screenshot. I'm new to node.js and js in general so I'd love some feedback if there's some workaround.
I still think that a change should be made here though, as empty catch blocks aren't a great practice.

sindresorhus:
Packages should not log directly to the console.

sindresorhus:
> I still think that a change should be made here though, as empty catch blocks aren't a great practice.

If you read the code, you'll see that not catching the error is intended.

sindresorhus:
Your problem is either a too old Node.js version or a JS parser not being able to handle the syntax here. Neither are problems with this package.

Issue #272, opened 2022-01-05
Apps or names cannnot try to open


masterX89:
## Basic Info

1. Windows

   - System: Windows 10 x64 (19043.1348)

   - Browser: 
     - Not have Firefox
     - Microsoft Edge 96.0.1054.62
     - Chrome 96.0.4664.110

2. MacOS
   - System: MacOS Monterey 12.0.1
   - Browser: 
     - Not have Firefox
     - Chrome 97.0.4692.71
     - Safari 15.1 (17612.2.9.1.20)

## What is expected?

> ##### app
>
> Type: 
>
> ```tsx
> {name: string | string[], arguments?: string[]} | Array<{name: string | string[], arguments: string[]}>
> ```
> Specify the `name` of the app to open the `target` with, and optionally, app `arguments`. `app` can be an array of apps to try to open and `name` can be an array of app names to try. If each app fails, the last error will be thrown.

## Steps to reproduce
I wrote the following code in the Windows which is consistent with the description in Basic Info. If the description of [app api](https://github.com/sindresorhus/open#app) in the [options](https://github.com/sindresorhus/open#options) is correct, then the web page will be opened by the chrome browser because there is no firefox browser. But the current situation is that the chrome browser or   the msedge browser does not open the web page.

```js
await open("https://google.com", {
    app: {
    	name: ["firefox", "chrome", "msedge"],
    },
});
```

```js
await open("https://google.com", {
    app: {
        name: ["firefox", "msedge", "chrome"],
    },
});
```

Then I wrote the similar code that change in the MacOS which is consistent with the description in Basic Info, just replaced `msedge` with `safari`. The browser still does not open the web page.

```js
await open("https://google.com", {
    app: {
    	name: ["firefox", "google chrome", "safari"],
    },
});
```

```js
await open("https://google.com", {
    app: {
        name: ["firefox", "safari", "google chrome"],
    },
});
```

## Analysis

After debugging, I think the problem lies in the `pTryEach` function:

```js
const pTryEach = async (array, mapper) => {
	let latestError;

	for (const item of array) {
		try {
			return await mapper(item); // eslint-disable-line no-await-in-loop
		} catch (error) {
			latestError = error;
		}
	}

	throw latestError;
};
```

The `mapper` function calls the `baseOpen` function, and no exception is thrown in the `baseOpen` function, so the `catch` is actually an unreachable statement.

[...]

PR #289, opened 2022-10-12
Wait 50ms for a process to error before returning

caass:
Potential fix for https://github.com/sindresorhus/open/issues/272, although it's ugly :/


sindresorhus:
I appreciate the PR, but don't think this is the correct fix.

PR #294, opened 2023-01-27
Implemented browser and browserPrivate

### LINKED ISSUE #266: To open url on default browser with incognito mode on cross platforms

leslieyip02:
Fixes #266 

Added `open.apps.browser` and `open.apps.browserPrivate`. 

`open.apps.browser` opens the default browser.
`open.apps.browserPrivate` opens the default browser in incognito mode for chrome, firefox and edge.

I tested this on Windows, but I'm not sure if it works on other platforms.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,25 @@ const localXdgOpenPath = path.join(__dirname, 'xdg-open');
 
 const {platform, arch} = process;
 
+// Podman detection
+const hasContainerEnv = () => {
+	try {
+		fs.statSync('/run/.containerenv');
+		return true;
+	} catch {
+		return false;
+	}
+};
+
+let cachedResult;
+function isInsideContainer() {
+	if (cachedResult === undefined) {
+		cachedResult = hasContainerEnv() || isDocker();
+	}
+
+	return cachedResult;
+}
+
 /**
 Get the mount point for fixed drives in WSL.
 
@@ -120,7 +139,7 @@ const baseOpen = async options => {
 		if (app) {
 			cliArguments.push('-a', app);
 		}
-	} else if (platform === 'win32' || (isWsl && !isDocker() && !app)) {
+	} else if (platform === 'win32' || (isWsl && !isInsideContainer() && !app)) {
 		const mountPoint = await getWslDrivesMountPoint();
 
 		command = isWsl ?
--- a/index.js
+++ b/index.js
@@ -1,10 +1,10 @@
-const path = require('path');
-const childProcess = require('child_process');
-const {promises: fs, constants: fsConstants} = require('fs');
-const isWsl = require('is-wsl');
-const isDocker = require('is-docker');
-const defineLazyProperty = require('define-lazy-prop');
-const defaultBrowser = require('x-default-browser');
+import path from 'path';
+import childProcess from 'child_process';
+import {promises as fs, constants as fsConstants} from 'fs';
+import isWsl from 'is-wsl';
+import isDocker from 'is-docker';
+import defineLazyProperty from 'define-lazy-prop';
+import defaultBrowser from 'default-browser';
 
 // Path to included `xdg-open`.
 const localXdgOpenPath = path.join(__dirname, 'xdg-open');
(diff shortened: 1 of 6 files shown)


sindresorhus:
Thanks for working on this.

I would prefer to use https://github.com/sindresorhus/default-browser

You also need to update the readme.

index.js
```diff
@@ -98,6 +101,37 @@ const baseOpen = async options => {
 		}));
 	}
 
+	if (app === 'browser') {
+		const browser = await defaultBrowser();
+		const browserName = browser.name.toLowerCase();
+		return baseOpen({
+			...options,
+			app: {
+				name: open.apps[browserName],
```
sindresorhus, line 110: You will need to do some normalisation of the names. `browserName` will be different depending on the OS.

index.js
```diff
@@ -309,7 +343,11 @@ defineLazyProperty(apps, 'edge', () => detectPlatformBinary({
...
 
+defineLazyProperty(apps, 'browser', () => 'browser');
+
+defineLazyProperty(apps, 'browserPrivate', () => 'browserPrivate');
+
 open.apps = apps;
 open.openApp = openApp;
 
-module.exports = open;
+export {open};
```
sindresorhus, line 353:
```suggestion
export default open;
```

package.json
```diff
@@ -50,7 +51,9 @@
 	"dependencies": {
 		"define-lazy-prop": "^2.0.0",
 		"is-docker": "^2.1.1",
-		"is-wsl": "^2.2.0"
+		"is-wsl": "^2.2.0",
+		"url": "^0.11.0",
```
sindresorhus, line 55: `url` is a built-in module.

index.js
```diff
@@ -309,7 +343,11 @@ defineLazyProperty(apps, 'edge', () => detectPlatformBinary({
 	wsl: '/mnt/c/Program Files (x86)/Microsoft/Edge/Application/msedge.exe'
 }));
 
+defineLazyProperty(apps, 'browser', () => 'browser');
+
+defineLazyProperty(apps, 'browserPrivate', () => 'browserPrivate');
+
 open.apps = apps;
 open.openApp = openApp;
```
sindresorhus, line 351: These should be named exports.

readme.md
```diff
@@ -121,20 +124,23 @@ We do not recommend setting this option. The convention for success is exit code
...
 ```
+`browser` and `browserPrivate` can also be used to access the user's default browser through [`default-browser`](https://github.com/sindresorhus/default-browser).
 
 #### Supported apps
 
 - [`chrome`](https://www.google.com/chrome) - Web browser
 - [`firefox`](https://www.mozilla.org/firefox) - Web browser
 - [`edge`](https://www.microsoft.com/edge) - Web browser
+- `browser` - Default web browser
+- `browserPrivate` - Default web browser in incognito mode
```
sindresorhus, line 143: This needs to document which browsers it supports.

sindresorhus (changes requested):

index.js
leslieyip02: You mean something like this?

``` js
export {apps};
export default open;
```

sindresorhus: Yes

index.js
```diff
@@ -123,13 +111,25 @@ const baseOpen = async options => {
...
+				return baseOpen({
+					...options,
+					app: {
+						name: open.apps[supportedBrowser],
+						arguments: appArguments
+					}
+				});
 			}
-		});
+		}
```
leslieyip02, line 130: I'm not too sure how to normalise the names, but is this good enough? It just checks if the `browserName` string contains `"chrome"`, `"firefox"` or `"edge"` since these are the only browsers supported right now.

sindresorhus: That is not going to work. What if the default browser is `Chrome Canary`?

sindresorhus:
Friendly bump

index.js
leslieyip02: Sorry for the lack of replies. I don't know what to do besides converting to lowercase and stripping whitespace. Do you have any advice for normalisation?

sindresorhus: A list of browsers with their mapping on the different operating systems. We only support 3 browsers for this, so should not be hard to make a list like that.

For example, `.chrome` is `Google Chrome` on macOS, `google-chrome` on Linux.

sindresorhus:
You still need to do this: https://github.com/sindresorhus/open/pull/294/files#r1093000444

sindresorhus:
Can you fix the merge conflict?

leslieyip02:
Hi @sindresorhus, would this be ok?

sindresorhus:
Thanks :)

PR #299, opened 2023-02-27
Fix systemroot inconsistent casing (Fixes #292)

### LINKED ISSUE #292: error - uncaughtException: Error: spawn undefined\System32\WindowsPowerShell\v1.0\powershell ENOENT
Windows 10

`open("https://localhost:3000")`

Codex-:
Long time no PR, sindresorhus 😄 

This PR simply adds a fallback for casing differences in `SYSTEMROOT` across windows versions, currently when using this on W11 (at least the 3 versions I tested) `SYSTEMROOT` was undefined, however `SystemRoot` _is_ defined.

So, try and use both

Fixes #292

### DIFF
--- a/index.js
+++ b/index.js
@@ -144,7 +144,7 @@ const baseOpen = async options => {
 
 		command = isWsl ?
 			`${mountPoint}c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe` :
-			`${process.env.SYSTEMROOT}\\System32\\WindowsPowerShell\\v1.0\\powershell`;
+			`${process.env.SYSTEMROOT || process.env.SystemRoot}\\System32\\WindowsPowerShell\\v1.0\\powershell`;
 
 		cliArguments.push(
 			'-NoProfile',


sindresorhus:
> On Windows operating systems, environment variables are case-insensitive.
> - https://nodejs.org/api/process.html#processenv

Codex-:
I read that too, however it doesn't seem to always be the case for all env vars:
image

```
PS C:\dev\bfp-app\packages\app> node -v    
v16.19.1
PS C:\dev\bfp-app\packages\app> npm -v     
9.5.1
```

sindresorhus:
Then open an issue on Node.js and add a code comment above your code change with a link to the Node.js issue. I prefer to only add workarounds when there's a chance it will be fixed properly at some point.

Codex-:
No problem, thanks anyway

sindresorhus:
Why are you closing?

Codex-:
> Why are you closing?

I'm unable to consistently reproduce this in a blank sandboxe, so I'm assuming that it must be some combination of our code and environment causing this as opposed to node or otherwise. If I find a more consistent reproduction in node, I'll re-open with an issue, until then I'm going to assume it's a me problem :)

PR #309, opened 2023-04-11
fix: remove __dirname

powerfulyang:
When I use this library in jest. "SyntaxError: Identifier '__dirname' has already been declared" has been thrown.


sindresorhus:
This is not a problem with this package.

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

PR #316, opened 2023-05-26
Fix passed an illegal parameter

cweijan:
Hi, if I pass parameters to openApp, the program will not recognize it, and it will work normally after removing the extra quotes.

### DIFF
--- a/index.js
+++ b/index.js
@@ -202,7 +202,7 @@ const baseOpen = async options => {
 		}
 
 		if (appArguments.length > 0) {
-			appArguments = appArguments.map(arg => `"\`"${arg}\`""`);
+			appArguments = appArguments.map(arg => `"${arg}"`);
 			encodedArguments.push('-ArgumentList', appArguments.join(','));
 		}
 


index.js
```diff
@@ -202,7 +202,7 @@ const baseOpen = async options => {
 		}
 
 		if (appArguments.length > 0) {
-			appArguments = appArguments.map(arg => `"\`"${arg}\`""`);
```
sindresorhus, line 205: It's double-quoted for a reason. You will have to prove why it's incorrect.

> // Double quote with double quotes to ensure the inner quotes are passed through.
			// Inner quotes are delimited for PowerShell interpretation with backticks.

sindresorhus:
// @tim-stasse

cweijan:
Hi, I am making this request because I am using mstsc and below is my code.
```js
import { openApp } from 'open';
const params = ['test.rdp','/v','127.0.0.1:3389']
openApp('mstsc.exe', { arguments: params });
```

image

After I remove the extra quotes, openApp will work fine.

Issue #318, opened 2023-07-24
Argument list quotes aren't escaped for PowerShell


ChristianLW:
When passing arguments (ending up in `options.app.arguments` for `baseOpen`), quotation marks in the passed string(s) aren't escaped with a <code>\`</code> for PowerShell when Windows is the current platform. This causes things like `--user-data-dir="foo"` to make the function silently fail.

PR #319, opened 2023-07-24
Escape quotes in arguments for PowerShell

ChristianLW:
A fix for #318. This is just a simple patch and can of course be changed if necessary.

The `String.prototype.replaceAll.call(arg, ...)` instead of just `arg.replaceAll(...)` is because `appArguments` might contain things other than strings.

This doesn't add escaping to the app name itself.

I'm not sure if you want to set up a unit test for this scenario, and I'm not sure if this change marks a breaking API change (if some people already put <code>\`</code> in their strings even though that wouldn't work on other platforms).

### DIFF
--- a/index.js
+++ b/index.js
@@ -202,7 +202,7 @@ const baseOpen = async options => {
 		}
 
 		if (appArguments.length > 0) {
-			appArguments = appArguments.map(arg => `"\`"${arg}\`""`);
+			appArguments = appArguments.map(arg => `"\`"${String.prototype.replaceAll.call(arg, '"', '`"')}\`""`);
 			encodedArguments.push('-ArgumentList', appArguments.join(','));
 		}
 
--- a/index.js
+++ b/index.js
@@ -202,7 +202,7 @@ const baseOpen = async options => {
 		}
 
 		if (appArguments.length > 0) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -202,7 +202,7 @@ const baseOpen = async options => {
 		}
 
 		if (appArguments.length > 0) {
-			appArguments = appArguments.map(arg => `"\`"${arg}\`""`);
+			appArguments = appArguments.map(arg => `"\`"${String.prototype.replaceAll.call(arg, '"', '`"')}\`""`);
```
sindresorhus, line 205:
```suggestion
			appArguments = appArguments.map(arg => `"\`"${String(arg).replaceAll('"', '`"')}\`""`);
```

sindresorhus: This does not handle the case where the quote is already escaped: `\"`

ChristianLW: Sorry for the late response. Could you please elaborate on the `\"` escaping? I can't find anything on using `\` as an escape character in PowerShell.

Alternatively, would it make sense to use single quote strings, as they don't interpret special characters? You can include single quotes in single quote strings by doubling them. This would almost certainly eliminate any edge cases.

Issue #205, opened 2020-11-11
Windows Command Prompt support broken


WillieRuemmele:
Hello, with this PR https://github.com/sindresorhus/open/pull/188 all of the command prompt users were broken. Is there a way this library could try to open with powershell and then, if that fails, revert to opening with command prompt?

sindresorhus:
How is it broken?

WillieRuemmele:
we, the [sfdx-cli](https://www.npmjs.com/package/sfdx-cli) have users who CANNOT install powershell on their computers and when they try to use one of our commands that relies on this library, their command blows up with `open spawn UNKNOWN.`. It seems like the linked PR above was when it was changed - I'm hoping to keep the current behavior but have it fall back to `cmd` if it fails

sindresorhus:
// @tim-stasse @Pytal

tabarra:
I'm not sure if that PR is the cause, but I can attest this update broke _things_.  
Had a few dozen people report this issue to me.  
  
Managed to grab the following screenshot:
![spawn undefined](https://i.imgur.com/GDMdcKZ.png)  
Unfortunately I do not have more information. For now I'm downgrading.

BillyHams:
I'm in the same situation.  Access to Powershell is blocked by Group Policies for me and all my teammates.  We are not 100% clear on how to downgrade easily.

DebuggingYou:
Hello together, 
are there any news about this issue? I have here the same issue on Windows 10 Pro (Version: 1809 / OS build: 17763.1397), and with my user I don't have access to PowerShell. 

I have received the following message from the SFDX CLI command, that might help? 
```
ERROR running auth:web:login: spawn UNKNOWN
*** Internal Diagnostic ***

Error: spawn UNKNOWN
at ChildProcess.spawn (internal/child_process.js:403:11)
at Object.spawn (child_process.js:553:9)
at module.exports (C:\Users\xxx\AppData\Roaming\npm\node_modules\sfdx-cli\node_modules\open\index.js:175:34)
at async Login.executeLoginFlow (C:\Users\xxx\AppData\Roaming\npm\node_modules\sfdx-cli\node_modules\@salesforce\plugin-auth\lib\commands\auth\web\login.js:54:9)
at async Login.run (C:\Users\xxx\AppData\Roaming\npm\node_modules\sfdx-cli\node_modules\@salesforce\plugin-auth\lib\commands\auth\web\login.js:35:30)
[...]

PR #328, opened 2023-12-22
fix(windows): add windir and C:/Windows as fallback on non-WSL

pikax:
related: https://github.com/vitejs/vite/issues/14292 https://github.com/vitejs/vite/discussions/14293

Not seeing `SYSTEMROOT` being provided on the environment on my system.

```bash  
System:
    OS: Windows 11 10.0.22621
    CPU: (32) x64 AMD Ryzen 9 7950X 16-Core Processor
    Memory: 35.64 GB / 63.14 GB
  Binaries:
    Node: 20.10.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.2.3 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.12.0 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Edge: Chromium (120.0.2210.77)
    Internet Explorer: 11.0.22621.1
```

`windir` seems to be providing the same path and just in case added the default `C:\\Windows` as a fallback.

### DIFF
--- a/index.js
+++ b/index.js
@@ -169,7 +169,7 @@ const baseOpen = async options => {
 
 		command = isWsl
 			? `${mountPoint}c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe`
-			: `${process.env.SYSTEMROOT}\\System32\\WindowsPowerShell\\v1.0\\powershell`;
+			: `${process.env.SYSTEMROOT || process.env.windir || 'C:\\Windows'}\\System32\\WindowsPowerShell\\v1.0\\powershell`;
 
 		cliArguments.push(
 			'-NoProfile',


sindresorhus:
The problem here is your system as `SystemRoot` is defined by default by the system, so you must have unset it somehow.

pikax:
> The problem here is your system as `SystemRoot` is defined by default by the system, so you must have unset it somehow.

That explanation is the one that makes more sense, but I don't recall changing/removing it, it could be removed by installing some software I reckon. 

Just for FYI, here's my current Environment variables, I believe the only thing I'll change be `Path`, the rest of the variables are either set by installing software or by system default.

image


Is a very odd behaviour that variable is not present,  regardless, this change should probably prevent that issue from happening since there's two fallbacks.

brc-dd:
Closes #205. But the core issue is probably with Node.js itself here as Sindre commented on #299.

@pikax Can you try logging `process.env.SystemRoot` instead of `process.env.SYSTEMROOT` in node? I don't think systemroot shows there in the settings - it's a readonly variable automatically added by windows.

pikax:
I can confirm `SystemRoot` is being provided instead of Uppercased

PR #337, opened 2024-03-08
Add support for Plasma (KDE) 6

### LINKED ISSUE #336: Nothing is opened under Plasma (KDE) 6
Plasma 6 was released about a week ago. When running this DE, nothing is opened (at least when trying to open an URL via a default browser, which is what I tested). If I understand it correctly, the problem is in the xdg-open shell script. If I have time, I'll try to create a fix, it should be simple enough.

ondrakucera:
Several notes worth mentioning:

- This should be good enough until Plasma 7 is released years from now (Plasma 6 was released roughly ten years after Plasma 5).
- However, it would be nicer if, in such a case, the script would fail (instead of doing nothing as it does now). I didn't want to go there but perhaps adding a "default" branch calling `exit_failure_operation_failed` would be the way to go?
- I'm expecting `kde-open` to be present under Plasma 6. This is the case on Arch Linux; I can't speak for other distributions. However, looking at how it is packaged on Arch, I don't see any weird customization, so I expect that if you build `kde-cli-tools` elsewhere, you will have `kde-open`, too.

Fixes #336.

### DIFF
--- a/xdg-open
+++ b/xdg-open
@@ -617,6 +617,9 @@ open_kde()
         5)
           kde-open${KDE_SESSION_VERSION} "$1"
         ;;
+        6)
+          kde-open "$1"
+        ;;
       esac
     else
         kfmclient exec "$1"


ondrakucera:
Now I see the script is actually taken from https://gitlab.freedesktop.org/xdg/xdg-utils/-/blob/master/scripts/xdg-open.in, which, in its current version, contains the same lines for Plasma 6 as my patch does. But it may also differ elsewhere, co it may make sense to just completely copy its newest version.

sindresorhus:
Yes, would be great if you could update it to the latest version.

ondrakucera:
Done (#338).

Issue #341, opened 2024-04-01
Older Node 18 versions do not provide a `constants` export


mkilpatrick:
Using Node 18.0.0 results in this error after upgrading from 9.x to 10.x.

```
import fs, {constants as fsConstants} from 'node:fs/promises';
            ^^^^^^^^^
SyntaxError: The requested module 'node:fs/promises' does not provide an export named 'constants'
```

It's an error until you use >= v18.4.0

wujingquan:
i got same problem in node v14, but node v16 and v18 no problem

luckfunc:
node14x: <img width="1027" alt="image" src="https://github.com/user-attachments/assets/d9c8d300-8d9b-4467-b0ec-230a1bfdd0c3">

node18: <img width="965" alt="image" src="https://github.com/user-attachments/assets/70aa75c3-0e75-4403-bb73-4ff52b8cb357">

kadinwu:
are you going to merge @luckfunc fix? i meet the same issue with Nodejs14 and upgrade nodejs is not easy.
Could you help fix it in priority? Thanks.

PR #345, opened 2024-09-08
fix: handle missing constants export in Node.js 14

luckfunc:
This PR addresses the issue where Node.js 14 does not export constants from 'fs/promises', causing compatibility problems. It checks the Node.js version and falls back to using 'fs.constants' when running in Node.js 14.

For more context, see(https://github.com/sindresorhus/open/issues/341)


sindresorhus:
This package requires Node.js 18.

luckfunc:
sorry for my question

PR #346, opened 2024-09-17
add `xdg-open` to module exports

fantua:
We bundle this package with `esbuild`. It will be usefull to add `xdg-open` to module exports so we can resolve it's path and copy to `build` folder.

Usage example:
```js
import path from 'node:path';
import { createRequire } from 'node:module';
import fs from 'node:fs/promises';

const require = createRequire(import.meta.url);

await fs.copyFile(
  require.resolve('open/xdg-open'),
  path.join('build', 'xdg-open')
);
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,11 @@
 	},
 	"type": "module",
 	"exports": {
-		"types": "./index.d.ts",
-		"default": "./index.js"
+		".": {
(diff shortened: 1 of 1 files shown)


sindresorhus:
This is the wrong approach. `exports` should only contain things that are meant for the public. The script is an implementation detail.

PR #349, opened 2024-10-15
feat: Prefer powershell core over powershell windows

ShreyasJejurkar:
This PR check does the user has new Powershell core, if it uses that, in case if it's not then it fallbacks to the old Windows tied PowerShell

### DIFF
--- a/index.js
+++ b/index.js
@@ -167,9 +167,15 @@ const baseOpen = async options => {
 	} else if (platform === 'win32' || (isWsl && !isInsideContainer() && !app)) {
 		const mountPoint = await getWslDrivesMountPoint();
 
+		// Check if PowerShell core (pwsh) is available
+		const pwshPath = `${process.env.USERPROFILE}\\AppData\\Local\\Microsoft\\WindowsApps\\pwsh.exe`;
+		const usePwsh = fs.existsSync(pwshPath);
+
 		command = isWsl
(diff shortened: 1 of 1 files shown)


sindresorhus:
Why does it matter which PowerShell it uses?

sindresorhus:
But that doesn't really matter in our case. We just use it to execute a single command. I can think of more downsides: potentially breaking changes, differing behavior, etc.

ShreyasJejurkar:
> potentially breaking changes, differing behavior

Powershell Core is the same as the old Powershell but with a better-performing runtime and the same feature parity as the old one. 

And also if we don't find a new PowerShell, we are anyway to fallback to the old one to avoid any breaking things.

sindresorhus:
> Powershell Core is the same as the old Powershell but with a better-performing runtime and the same feature parity as the old one.

Minor performance differences does not really matter much here. The slow part is launching the apps, not executing the command. And you haven't really provided any evidence of performance improvements for this specific case.

> And also if we don't find a new PowerShell, we are anyway to fallback to the old one to avoid any breaking things.

That's not the kind of breaking changes I'm talking about. The core one is a moving target. It may be identical today, but in 2 years, it may break some feature or subtly change some behavior.

sindresorhus:
I'm going to pass on this as the benefit does not outweigh the risk.

PR #350, opened 2024-11-12
fix: support Single Executable Applications

sebamarynissen:
I'm using this module with Node's new [Single Executable Application functionality](https://nodejs.org/api/single-executable-applications.html). Currently this requires you to bundle your application into a single JavaScript file, which I'm doing with esbuild. However, when importing the module, it chokes over
```js
const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
because after bundling with esbuild, `import.meta.url` is undefined.

This PR fixes that by falling back to `process.execPath` in case `import.meta.url` is `undefined`. This also means that if someone wants to use the bundled xdg binary, they can distribute it alongside the sea binary and it should work.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,7 +10,8 @@ import defaultBrowser from 'default-browser';
 import isInsideContainer from 'is-inside-container';
 
 // Path to included `xdg-open`.
-const __dirname = path.dirname(fileURLToPath(import.meta.url || process.execPath));
+const self = import.meta.url;
+const __dirname = path.dirname(self ? fileURLToPath(self) : process.execPath);
 const localXdgOpenPath = path.join(__dirname, 'xdg-open');
 
 const {platform, arch} = process;
--- a/index.js
+++ b/index.js
@@ -10,7 +10,7 @@ import defaultBrowser from 'default-browser';
 import isInsideContainer from 'is-inside-container';
 
 // Path to included `xdg-open`.
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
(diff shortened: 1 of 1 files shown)


reneleonhardt:
@sindresorhus Are there objections against this improvement?

sindresorhus:
Thanks for the PR, but I don't think falling back to `process.execPath` is correct.

`import.meta.url` points to the module file, while `process.execPath` points to the Node.js/SEA executable. Using `path.dirname(process.execPath)` would look for `xdg-open` in the executable's directory (e.g., `/usr/local/bin/`), not the `open` package directory.

This is also a security issue - if the executable is in a user-writable directory, an attacker could place a malicious `xdg-open` binary there.

The code already handles bundled scenarios at line 214. When `import.meta.url` is undefined, `__dirname` will be undefined, triggering the fallback to system `xdg-open`.

Note: [Node.js SEA only supports CommonJS](https://nodejs.org/api/single-executable-applications.html), not ESM. For bundling `xdg-open` with SEA, use the [assets feature](https://nodejs.org/api/single-executable-applications.html#generating-single-executable-preparation-blobs).

reneleonhardt:
@sebamarynissen Did you try the official workarounds with `--define` or writing a plugin?

https://www.perplexity.ai/search/nodejs-single-executable-appli-JGuE6P6_SLauNTHDK01NqA#0
https://github.com/evanw/esbuild/issues/1492#issuecomment-893144483
https://github.com/evanw/esbuild/issues/2249#issuecomment-1125873862

Issue #357, opened 2025-04-29
Cannot detect default browser correctly when app.name is set to browser or browserPrivate in WSL environment.


sukjuhong:
Currently, when `app.name` is set to browser or browserPrivate, the default-browser library is used to detect the system's default browser.

```js
// in 'default-browser' library
if (process.platform === 'linux') {
	const {stdout} = await execFileAsync('xdg-mime', ['query', 'default', 'x-scheme-handler/http']);
	const id = stdout.trim();
	const name = titleize(id.replace(/.desktop$/, '').replace('-', ' '));
	return {name, id};
}
```

Since WSL reports process.platform as linux, it uses xdg-mime to detect the default browser.

However, in WSL, the Windows-installed default browser is usually not set properly for x-scheme-handler/http, leading to incorrect or missing results.

In my testing with a fresh Ubuntu installation on WSL:
- x-scheme-handler/http is often unset.
- If wslu is installed, the result might be { name: 'Wslview', id: 'wslview.desktop' }.
- Sometimes it even defaults to Firefox even when Firefox isn't installed.

This leads to runtime errors like:
```log
Error: Wslview is not supported as a default browser
    at baseOpen (file:///home/***/dev/test-open-browser/node_modules/open/index.js:143:9) 
    at async file:///home/***/dev/test-open-browser/index.js:4:1
```
Since Wslview isn't an actual browser, the open operation fails.

```js
import open from "open";

await open("https://google.com", { wait: true });               // Works fine
await open("https://google.com", { app: { name: "browser" } }); // Fails
```

When `app.name` is not set, it falls back to using PowerShell's Start command, and the URL correctly opens in the default browser.

However, to use features like private/incognito mode, setting `app.name` is necessary.

Is there a good way to handle this situation in WSL?

Maybe it would be possible to detect WSL specifically and fall back to the Windows behavior instead of using xdg-mime? Any advice would be appreciated!

PR #358, opened 2025-04-30
feat: add function to detect Windows default browser from WSL

sukjuhong:
related issue: #357

To provide additional explanation on this issue, I found that in order for the user to change `x-scheme-handler/http`, the actual .desktop file must exist.
To forcefully change it, you would need to modify `/usr/share/applications/defaults.list`, but I thought this wasn't the common behavior.
Therefore, I wrote a function to get the default browser in Windows using a PowerShell command, and linked it with the `browser.id` value, so that if `isWsl` is true, the function would be executed.

## Changes

- Added getWindowsDefaultBrowser() to retrieve the default Windows browser from WSL
- Maps ProgId (e.g. ChromeHTML) to existing IDs

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,12 +2,15 @@ import process from 'node:process';
 import {Buffer} from 'node:buffer';
 import path from 'node:path';
 import {fileURLToPath} from 'node:url';
-import childProcess, {execFile} from 'node:child_process';
+import util from 'node:util';
+import childProcess from 'node:child_process';
 import fs, {constants as fsConstants} from 'node:fs/promises';
 import isWsl from 'is-wsl';
 import defineLazyProperty from 'define-lazy-prop';
 import defaultBrowser from 'default-browser';
 import isInsideContainer from 'is-inside-container';
+
+const execFile = util.promisify(childProcess.execFile);
 
 // Path to included `xdg-open`.
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
@@ -61,12 +64,20 @@ const getWslDrivesMountPoint = (() => {
 })();
 
 /**
+ * Get the PowerShell executable path in WSL environment.
+ * @returns {Promise<string>} The full path to the PowerShell executable in WSL
+ */
+const getPowershellPathFromWsl = async () => {
+	const mountPoint = await getWslDrivesMountPoint();
+	return `${mountPoint}c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe`;
+};
+
+/**
  * Get the default browser name in Windows from WSL.
  * @returns {Promise<string>} Browser name
  */
 async function getWindowsDefaultBrowserFromWsl() {
-	const mountPoint = await getWslDrivesMountPoint();
-	const powershellPath = `${mountPoint}c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe`;
+	const powershellPath = await getPowershellPathFromWsl();
 
 	const rawCommand
 		= `$prog = Get-ItemProperty -Path 'HKCU:\\Software\\Microsoft\\Windows\\Shell\\Associations\\UrlAssociations\\http\\UserChoice' | Select-Object -ExpandProperty ProgId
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -60,6 +60,56 @@ const getWslDrivesMountPoint = (() => {
...
+	const powershellPath = `${mountPoint}c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe`;
+
+	const rawCommand
+		= `$prog = Get-ItemProperty -Path 'HKCU:\\Software\\Microsoft\\Windows\\Shell\\Associations\\UrlAssociations\\http\\UserChoice' | Select-Object -ExpandProperty ProgId
+	Write-Output $prog
+	`.trim();
+	const encodedCommand = Buffer.from(rawCommand, 'utf16le').toString('base64');
+
+	return new Promise((resolve, reject) => {
+		execFile(
```
sindresorhus, line 78: Use the promise version: https://nodejs.org/api/child_process.html#:~:text=If%20this%20method%20is%20invoked%20as%20its%20util.promisify()ed%20ver,-sion

index.js
```diff
@@ -60,6 +60,56 @@ const getWslDrivesMountPoint = (() => {
 	};
 })();
 
+/**
+ * Get the default browser name in Windows from WSL.
+ * @returns {Promise<string>} Browser name
+ */
+async function getWindowsDefaultBrowserFromWsl() {
+	const mountPoint = await getWslDrivesMountPoint();
+	const powershellPath = `${mountPoint}c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe`;
```
sindresorhus, line 69: Reuse the logic with: https://github.com/sindresorhus/open/blob/9c73b0c1b8e07dc153f7f0b140a6210e3c27c8ea/index.js#L222-L223

index.js
```diff
@@ -2,7 +2,7 @@ import process from 'node:process';
 import {Buffer} from 'node:buffer';
 import path from 'node:path';
 import {fileURLToPath} from 'node:url';
-import childProcess from 'node:child_process';
+import childProcess, {execFile} from 'node:child_process';
```
sindresorhus, line 5: You can just access `execFile` on `childProcess`.

sindresorhus (changes requested):

index.js
sukjuhong: Would it be appropriate to implement what you mentioned as a function like the one below, such as `getPowershellPath`?

```js
const getPowershellPath = (mountPoint) => { ... }
```

sindresorhus: Yes, I don't think (however, I haven't looked at the code thoroughly) it needs to accept `mountPoint`, it could just call `const mountPoint = await getWslDrivesMountPoint();` internally.

sukjuhong:
@sindresorhus

Thank you for the review! 🙌
I’ve addressed all the requested changes.
Please have a look and let me know if anything else needs to be adjusted.

index.js
```diff
@@ -60,6 +63,57 @@ const getWslDrivesMountPoint = (() => {
...
+			'-NonInteractive',
+			'-ExecutionPolicy',
+			'Bypass',
+			'-EncodedCommand',
+			encodedCommand,
+		],
+		{encoding: 'utf8'},
+	);
+
+	if (stderr && !stderr.includes('CLIXML')) {
```
sindresorhus, line 103: Can you add a short code comment about the `CLIXML` check? Not clear what it's about.

index.js
```diff
@@ -60,6 +63,59 @@ const getWslDrivesMountPoint = (() => {
...
+Get the default browser name in Windows from WSL.
+
+@returns {Promise<string>} Browser name.
+*/
+async function getWindowsDefaultBrowserFromWsl() {
+	const powershellPath = await getPowershellPathFromWsl();
+
+	const rawCommand = `$prog = Get-ItemProperty -Path 'HKCU:\\Software\\Microsoft\\Windows\\Shell\\Associations\\UrlAssociations\\http\\UserChoice' | Select-Object -ExpandProperty ProgId
+	Write-Output $prog
+	`.trim();
```
sindresorhus, line 86: Does it work if we do:

```suggestion
	const rawCommand = `
$prog = Get-ItemProperty -Path 'HKCU:\\Software\\Microsoft\\Windows\\Shell\\Associations\\UrlAssociations\\http\\UserChoice' | Select-Object -ExpandProperty ProgId
Write-Output $prog
		`.trim();
```

That would be the most readable.

sindresorhus: I cannot test this myself, but AI said it could be simplified to:

```js
const rawCommand = '(Get-ItemProperty -Path "HKCU:\\Software\\Microsoft\\Windows\\Shell\\Associations\\UrlAssociations\\http\\UserChoice").ProgId';
```

sindresorhus (changes requested):

index.js
sukjuhong: To be honest, I don’t know why `stderr` is being printed even when the response is normal.
According to ChatGPT, using `promisify` can capture and print content even when there’s no actual error.
In fact, when I print out the contents of `stderr`, the part where an error should appear actually includes the word "Completed".

```xml
 #< CLIXML
<Objs xmlns="http://schemas.microsoft.com/powershell/2004/04" Version="1.1.0.1">
  <Obj S="progress" RefId="0">
    <TN RefId="0">
      <T>System.Management.Automation.PSCustomObject</T>
      <T>System.Object</T>
    </TN>
    <MS>
      <I64 N="SourceId">1</I64>
      <PR N="Record">
        <AV>Preparing modules for first use.</AV>
        <AI>0</AI>
        <Nil/>
        <PI>-1</PI>
        <PC>-1</PC>
        <T>Completed</T> << here
        <SR>-1</SR>
        <SD> </SD>
      </PR>
    </MS>
  </Obj>
  <Obj S="progress" RefId="1">
    <TNRef RefId="0"/>
    <MS>
      <I64 N="SourceId">1</I64>
      <PR N="Record">
        <AV>Preparing modules for first use.</AV>
        <AI>0</AI>
        <Nil/>
        <PI>-1</PI>
        <PC>-1</PC>
        <T>Completed</T> << here
        <SR>-1</SR>
        <SD> </SD>
      </PR>
    </MS>
  </Obj>
</Objs>
```

How about ignoring stderr and just using stdout?
If a real error occurs, an error object would be thrown at that point anyway.

index.js
sukjuhong: I tested this command, and it works well!
I’ll go ahead and switch to using this one.

index.js
sindresorhus: > How about ignoring stderr and just using stdout?

👍

sukjuhong:
@sindresorhus
I’ve removed the stderr and tested to ensure it’s working properly.

PR #370, opened 2026-05-22
Detach non-wait PowerShell launches

### LINKED ISSUE #367: Can't open a local file anymore in v11
Hi,

I've been using this library successfully for quite some time to open a local file (the "index.html" file containing a Jest coverage report). Unfortunately, since the update to v11, it no longer works. Is this a breaking change (I haven't been able to find the changelog)? Or is it a bug?

I am on Windows 10.

Thanks.

puneetdixit200:
Summary:
- detach non-wait PowerShell launches after stdio is ignored so Windows targets can keep opening after the parent process exits
- add an AVA regression test that loads the module with `process.platform` mocked to `win32` and verifies the spawn options

Tests:
- `npm_config_cache=/tmp/open-npm-cache npx ava test.js --match="detaches PowerShell launches when not waiting on Windows"`
- `npm_config_cache=/tmp/open-npm-cache npx ava test.js --match="app launches resolve before close without fallback" --match="detaches PowerShell launches when not waiting on Windows"`
- `npm_config_cache=/tmp/open-npm-cache npm test`
- `git diff --check`

Fixes #367

### DIFF
--- a/index.js
+++ b/index.js
@@ -208,6 +208,7 @@ const baseOpen = async options => {
 			// PowerShell will keep the parent process alive unless stdio is ignored.
 			childProcessOptions.stdio = 'ignore';
 			childProcessOptions.detached = true;
+			childProcessOptions.windowsHide = true;
 		}
 	} else {
 		if (app) {
--- a/index.js
+++ b/index.js
@@ -207,6 +207,7 @@ const baseOpen = async options => {
 		if (!options.wait) {
 			// PowerShell will keep the parent process alive unless stdio is ignored.
 			childProcessOptions.stdio = 'ignore';
+			childProcessOptions.detached = true;
 		}
 	} else {
 		if (app) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Could you also set `windowsHide: true` when using `detached: true` here?

I double checked the Node docs, and on Windows a detached child process gets its own console window, while `windowsHide` defaults to `false`. Since this is the normal non-wait `open()` path, this could make a PowerShell window flash or show up for every call.

And make sure you test manually in PowerShell after this to confirm it works fine. I want to see a screenshot of it too, to confirm, since this PR is clearly AI generated and I don't really trust it.

puneetdixit200:
Updated in `6b71ed2`: the non-wait PowerShell path now sets `windowsHide: true` together with `detached: true` and `stdio: "ignore"`, and the regression test asserts those spawn options.

Local validation run on macOS:
- `npm install --package-lock=false --cache /tmp/open-370-npm-cache --no-audit`
- `npm test`
- `npm exec ava -- test.js -m "detaches PowerShell launches when not waiting on Windows"`
- `git diff --check`

I cannot provide the requested PowerShell screenshot from this macOS environment; that manual Windows validation still needs to be done on a Windows machine.

sindresorhus:
As I expected, just AI slop.

Issue #189, opened 2020-08-10
process.exit kills browser?


axi92:
With this code it is not working:
```
const open = require('open');

open('google.com');
process.exit(0);
```
If I remove the exit it is working. But I need an exit in my node script after the open() is called.

Issue #144, opened 2019-07-15
Impossible to trap errors from child process


heyimalex:
Howdy! This is an explanation for #56

For various reasons, cmd isn't in always in the system path on windows. This makes calls to spawn fail with an `ENOENT`.

This wouldn't be an issue on its own, except that `open` gives no way to trap that error. Because the `open` function is async, by the time the promise resolves and we have access to a ChildProcess that we can add an error event listener to, the error event has already fired and bubbled up to the top level, crashing the program with an `ENOENT`.

atlefren:
Some investigation:

Given an environment with no cmd (as described above)

edit index.js (https://github.com/sindresorhus/open/blob/master/index.js) line 131 and onwards:

```
	subprocess.unref();

	subprocess.on('error', function( err) {
		console.log('error from subprocess')
	});
	console.log("return subprocess")
	return subprocess;
```

Then import index.js in another file and run:

```
var options = { app: browser, wait: false };
open(url, options).then(process => {
      console.log("got process")
      process.on('error', function( err) {
        console.log('got error')
      })
      
    }).catch((e) => {
      console.log(e)
    });
```
This will print:

```
return subprocess
error from subprocess
got process
```

This means that the subprocess error is emitted before the caller gets the process. This means that we will not be able to listen for the error event in the caller.  This also means that we cannot throw an error in the event callback in open, as this will fire too late.

One possibility is to force a wait, this seems to work. The problem here is that when things work out, we will be waiting for the process to exit for a loooong time..

One (hacky?) solution is to always "wait", but when not wait is set, use a setTimeout to resolve if things does not go wrong after a (short) while, someting like this seems to wotk

```
return new Promise((resolve, reject) => {
		subprocess.once('error', reject);

		subprocess.once('close', exitCode => {
			if (exitCode > 0) {
				reject(new Error(`Exited with code ${exitCode}`));
				return;
			}

			if (options.wait) {
				resolve(subprocess);
			}

		});
		if (!options.wait) {
			setTimeout(() => {	resolve(subprocess);}, 100);
		}
	});
```

heyimalex:
Some other solutions:

- Look up the executable in the path manually before calling spawn. This is actually how it's handled in go's stdlib `os/exec` package, so there's some precedence here. However, doesn't prevent the broader issue of attaching error listeners to a process returned from an async function.
- Allow a callback option that's immediately called with the created ChildProcess so that we can synchronously attach listeners. This is kind of ugly but gets the job done.
[...]

PR #375, opened 2026-09-04
Fix Windows launcher race condition

### LINKED ISSUE #298: Issue: Windows not opening browser or any client application without `wait: true`
Hi

I am currently having an issue when using `open` when simply opening a URL on Windows. I use the following code:
```
const open = require('open');
const result = await open('https://google.be/');
```
This will run the code, but the browser will not open. I have noticed that, when the code takes longer to execute, by for example adding a for loop that takes +- 5 seconds to run, that the browser will in fact open and it will work.

I have also noticed, that by adding `{ wait: true }`, it will also start working, but that is not an option because when testing it on my macbook it will explicitly wait until the entire Chrome process has closed.

Is there any chance that you could have a look at this?

Thanks!
Mathijs

otnc:
## Problem

On Windows, `open()` (without `wait: true`) can silently fail to open anything, even though it resolves without throwing.

This reproduces with a plain call like:

```js
import open from 'open';

await open(String.raw`C:\path\to\file.html`, {app: {name: 'brave'}});
console.log('done');
// -> resolves without error, but the browser never opens
```

This is not specific to any particular runner (npm scripts, `npx`, etc.) — it reproduces with a bare `node` invocation that does nothing after the `await`.

## Root cause

On Windows, the actual command is always wrapped in a `powershell.exe` launcher that runs a base64-encoded `Start-Process ...` script. The comment above the non-fallback resolution path assumes:

> The launcher (open/xdg-open/PowerShell) exits quickly (~10-30ms) even on success.

That assumption holds for macOS `open` and `xdg-open`, but not for legacy Windows PowerShell (`System32\WindowsPowerShell\v1.0\powershell.exe`, which `powerShellPath()` always targets — Windows still isn't defaulting new processes to `pwsh`, so this is the common case, not an edge case). Starting it up (loading the runtime, applying `-ExecutionPolicy Bypass`, decoding and parsing the script) routinely takes several hundred ms, well past that ~10-30ms assumption.

The non-fallback, non-`wait` path only waits for the `'spawn'` event before resolving:

```js
subprocess.once('spawn', () => {
	subprocess.off('error', reject);
	resolve(subprocess);
});
```

`'spawn'` only means the OS handed back a process handle for `powershell.exe` — not that it has finished executing the encoded `Start-Process` command yet. If the caller does nothing meaningful after `await open(...)` (a very common pattern, e.g. a short-lived CLI script), the process can exit right after this resolves. Since the Windows branch never sets `detached: true` on `childProcessOptions` (only the Linux `xdg-open` branch does), the still-starting `powershell.exe` child is not detached from the caller, and can be torn down before it ever gets to run `Start-Process`.

I confirmed this directly: replacing the `spawn`-based resolution with one that waits for the launcher's `'close'` event (what the existing `isFallbackAttempt` path already does) makes it succeed reliably, taking ~200-900ms depending on system load — consistent with legacy PowerShell startup cost, not with the app itself.

Related: #298 (closed as "probably fixed" by 966239c, which added the `'spawn'` wait — that made things better but didn't fully close the race on Windows specifically, for the reason above). Also related: #144, #189.

## Fix

Reuse the existing `isFallbackAttempt` handling (which already waits for the launcher's `'close'` event before resolving) for the Windows path too, since Windows always goes through this same kind of short-lived launcher process. This does **not** wait for the *opened app* to close — only for the (still fast, just not ~10-30ms fast) launcher script that `Start-Process` runs in. It's unrelated to `options.wait`, which additionally passes `-Wait` to `Start-Process` itself to block until the opened app exits.

## Testing

- `xo` passes on the changed files.
- Added a Windows-specific `ava` test mirroring the existing `'app launches resolve before close without fallback'` test, but asserting the opposite (resolves *after* close) — gated by `process.platform === 'win32'`, leaving the non-Windows test and behavior untouched.
- Manually verified against a real `brave` launch on Windows 11: without this change, the browser did not open; with it, `open()` waits for the launcher and the browser reliably opens.

Fixes #298

### DIFF
--- a/index.js
+++ b/index.js
@@ -280,7 +280,18 @@ const baseOpen = async options => {
 	// When we're in a fallback attempt, we need to detect launch failures before trying the next app.
 	// Wait for the close event to check the exit code before unreffing.
 	// The launcher (open/xdg-open/PowerShell) exits quickly (~10-30ms) even on success.
-	if (isFallbackAttempt) {
+	//
+	// On Windows, the actual command always runs inside a `powershell.exe` launcher (see
+	// above). Unlike `open`/`xdg-open`, this launcher can take noticeably longer than
+	// ~10-30ms to start executing the encoded script, so waiting only for `'spawn'` isn't
+	// enough: the promise can resolve, and the caller's process can exit, before
+	// `Start-Process` has actually run. Since the launcher isn't detached on Windows,
+	// exiting at that point can tear it down before it launches the target app. Always
+	// wait for the launcher's own `'close'` event on Windows to avoid this race. This does
+	// not wait for the *opened app* to close, only for the (fast) launcher script.
+	const isWindowsLauncher = platform === 'win32' || shouldUseWindowsInWsl;
+
+	if (isFallbackAttempt || isWindowsLauncher) {
 		return new Promise((resolve, reject) => {
 			subprocess.once('error', reject);
 
--- a/test.js
+++ b/test.js
@@ -139,13 +139,9 @@ test('subprocess is spawned before promise resolves', async t => {
 	t.true(childProcess.pid !== undefined && childProcess.pid !== null);
 });
 
-test.serial('app launches resolve before close without fallback', async t => {
+const stubSpawnEmittingCloseAfter = delay => {
 	const originalSpawn = childProcess.spawn;
-	t.teardown(() => {
-		childProcess.spawn = originalSpawn;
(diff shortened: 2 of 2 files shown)


sindresorhus:
Closing in favor of https://github.com/sindresorhus/open/commit/734b821c36ee959dc11e380c3202770969347274

